PR #217, opened 2022-09-04
Update data quality metric outputs

grgmiller:
As part of the research paper for OGEI, I wanted to update some of the output data quality metrics and figured these changes would be useful to merge for our next release.

This PR makes the following changes:

- Combines `diba_imputation_performance.csv` and `national_imputation_performance.csv` into a single file `wind_solar_profile_imputation_performance.csv` which contains one column for the DIBA results and another column for the national results.
- Adds a new output `cems_pollutant_measurement_quality.csv` which summarizes what percentage of CO2 mass, SO2 mass, and NOx mass reported in CEMS was directly measured versus imputed. This metric is based on the reported "mass_measurement_code" reported for each observation in the CEMS data.
- Adds additional summary columns to `input_data_source.csv` and `hourly_profile_method.csv` for CO2e, SO2, and NOx (both totals and _for_electricity values)
- Breaks out the EIA column  in the `input_data_source.csv` to distinguish between monthly-reported EIA data and annually reported EIA data. 
- Re-orders certain rows in the outputs so that they are generally arranged with "better" sources/methods at the top and "worse" ones at the bottom.

**NOTE**: I am requesting this to be merged into a new `development` branch where we can stage cumulative updates before releasing a new version on `main`


gailin-p (commented):
Can you confirm that this works with `data_pipeline.py --small True`? Some previous per-method data quality metrics failed because the `--small` run doesn't use all of the methods, so just want to confirm that this avoids those issues

grgmiller:
I have confirmed that this works with `--small` in 2020

gailin-p (approved):

PR #221, opened 2022-09-07
Gailin/fix consumed sign bug

### LINKED ISSUE #214: Negative EFs for some BAs
Perhaps this can be explained by a more detailed reading of the OGEI methodology; some BAs have negative EFs. For example, for the annual 2020 data, SCL, PGE, and PSEI all have negative values. I would expect the minimum EF to be 0, which would represent a 100% renewable mix. Are these negative values errors? If not, would an entity really be able to report negative emissions from their use of carbon-free electricity procured from one of these BAs, or should the EFs be rounded up to 0?

gailin-p:
Closes #214 , see issue for more commentary

* Negative generated emission rates are set to zero during `write_power_sector_results`
* Negative consumed emission rates were bugs, these were fixed by: 
  * fixing the incorrect sign on interchange during calculation of `net_consumed_mwh`
  * using EIA-930 demand instead of `net_consumed_mwh` for BAs with inconsistencies between EIA-930 interchange and our `net_generation_mwh`, because these inconsistencies prevent calculation of a reasonable `net_consumed_mwh`
* Add rate columns to validation check for negative values 
* Add running of validation check to `output_to_results` to guarantee that all result files must pass validation check


.gitignore
```diff
@@ -14,6 +14,9 @@ notebooks/.ipynb_checkpoints
 notebooks/*/.ipynb_checkpoints
 .hypothesis/
 
+# Editor settings
+.vscode/*
```
grgmiller, line 18: This seems to duplicate line 24 (although I hadn't included the wildcard) - maybe we should just delete line 24

src/consumed.py
```diff
@@ -280,9 +277,14 @@ def output_results(self):
         for ba in self.regions:
             if (ba in self.import_regions) or (ba in self.generation_regions):
                 continue
-            self.results[ba]["net_consumed_mwh"] = (
-                self.generation[ba] + self.eia930.df[KEYS["E"]["TI"] % ba]
-            )[self.generation.index]
+            if ba in BA_930_INCONSISTENCY:
```
grgmiller, line 280: Great, I like how we only use demand for these BAs that have issues.

src/output_data.py
```diff
@@ -77,13 +92,18 @@ def output_intermediate_data(df, file_name, path_prefix, year, skip_outputs):
 
 
 def output_to_results(df, file_name, subfolder, path_prefix, skip_outputs):
+    # Always check columns that should not be negative.
+    small = "small" in path_prefix
```
grgmiller, line 96: This is a clever way to do this without passing args.small to the function!

Issue #220, opened 2022-09-07
Re-assess cleaning of EIA-930 data


gailin-p:
# Physics-based cleaning 

We use `gridemissions` to clean EIA 930 data. Currently we use 930 generation, however, this causes issues during the calculation of consumed emission rates. (see #214)

A better solution may be to do physics-based cleaning using our own generated MWh. This would require modifying the optimization algorithm to prioritize changing interchange and demand values over net generation values, since we trust our net generation estimates over the 930 values. 

We would also need to provide user guidance on compiling `cvxpy`, since it can be extremely slow for some users and this change would require running optimization during every run of `data_pipeline.py`. 

# Systematic issues 

The initial correction of systematic issues in 930 data depended on [comparison of interchanges between BAs](https://singularity-docs.stoplight.io/docs/open-grid-emissions-docs/cleaning_eia930-eia-930-data#systematic-data-issues). Now that we have the OGE dataset, we can do a more thorough comparison. 

One example of this is AZPS, where there is a sign flip on June 1, 2020, that was not caught in the initial analysis. 

<img width="1129" alt="Screen Shot 2022-09-06 at 9 47 37 PM" src="https://user-images.githubusercontent.com/12755256/188770751-54605dfc-a1e4-440b-a62d-1acc2dc42d7e.png">

grgmiller:
So for this specific issue with AZPS, it looks like in the raw data, the AZPS-SRP interchange flips from positive (net exporter to SRP) to negative (net importer from SRP) on Jun 1, 2020. 
![image](https://user-images.githubusercontent.com/45949268/188925194-1946ae2d-f1fc-4160-90a6-2088c73f12cc.png)

However, SRP reports being a net exporter to AZPS consistently across this period
![image](https://user-images.githubusercontent.com/45949268/188925517-3a75931d-8168-43da-8cf9-7db02d07bd7a.png)

Thus I think that AZPS had been incorrectly reporting its imports from SRP prior to 6/1/2020 as exports. They fixed it going forward, but never corrected the historical data. To fix this issue we would want to flip the sign of the AZPS-SRP interchange prior to this date.

grgmiller:
It almost appears that if you subtracted ~3000 from the SRP numbers, then flipped it, you would get approximately what AZPS was reporting...

grgmiller:
[...]

PR #222, opened 2022-09-07
Gailin/azps patch

gailin-p:
Advances #220 with an AZPS-SRP fix. Because AZPS-SRP has multiple different and not straightforward problems (see #220), we use SRP-AZPS as the ground truth. 

Also includes code to add data quality metrics to zipped output, change naming convention of zipped files for easier upload. 

Also includes small fix to `validate_vs_egrid` notebook (column name change from `plant_name` to `plant_name_eia`)

### DIFF
--- a/src/eia930.py
+++ b/src/eia930.py
@@ -441,12 +441,19 @@ def manual_930_adjust(raw: pd.DataFrame):
     ovec_col = get_int_columns("PJM", raw.columns, ["OVEC"])
     raw.loc[:, ovec_col] = raw.loc[:, ovec_col] * -1
 
-    # Interchange AZPS - SRP is wonky. Use SRP - AZPS (inverted)
+    # Interchange AZPS - SRP is wonky before 6/1/2020 7:00 UTC. Use SRP - AZPS (inverted)
     azps_srp = get_int_columns("AZPS", raw.columns, ["SRP"])
     srp_azps = get_int_columns("SRP", raw.columns, ["AZPS"])
-    raw.loc[:, azps_srp] = (raw.loc[:, srp_azps] * (-1)).rename(
+    replacement = (raw.loc[:, srp_azps] * (-1)).rename(
         columns={srp_azps[0]: azps_srp[0]}  # rename so Pandas will do the right thing
     )
+    raw.loc[raw.index < "2020-06-01T07:00+00", azps_srp] = replacement[
+        raw.index < "2020-06-01T07:00+00"
+    ]
+    # Update total interchange
+    all_cols = [c for c in get_int_columns("AZPS", raw.columns) if "ALL" not in c]
+    total_col = "EBA.AZPS-ALL.TI.H"
+    raw.loc[:, total_col] = raw.loc[:, all_cols].sum(axis=1)
 
     # Interchange TEPC is uniformly lagged
     cols = get_int_columns("TEPC", raw.columns)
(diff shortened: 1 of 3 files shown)


src/eia930.py
```diff
@@ -401,6 +401,10 @@ def manual_930_adjust(raw: pd.DataFrame):
             - PJM-OVEC, all time. Based on OVEC demand - generation, OVEC
                 should be a net exporter to PJM
                 - Note: OVEC's data repeats daily starting in 2018...
+        - Interchange mysterious
```
grgmiller, line 404: Haha love the category

src/output_data.py
```diff
@@ -49,7 +49,7 @@ def prepare_files_for_upload(years):
 
     This should only be run when releasing a new minor or major version of the repo.
     """
-    zip_data_for_zenodo()
+    # zip_data_for_zenodo()
```
grgmiller, line 52: Should this be uncommented?

src/eia930.py
```diff
@@ -437,6 +441,13 @@ def manual_930_adjust(raw: pd.DataFrame):
     ovec_col = get_int_columns("PJM", raw.columns, ["OVEC"])
     raw.loc[:, ovec_col] = raw.loc[:, ovec_col] * -1
 
+    # Interchange AZPS - SRP is wonky. Use SRP - AZPS (inverted)
+    azps_srp = get_int_columns("AZPS", raw.columns, ["SRP"])
+    srp_azps = get_int_columns("SRP", raw.columns, ["AZPS"])
+    raw.loc[:, azps_srp] = (raw.loc[:, srp_azps] * (-1)).rename(
```
grgmiller, line 447: We might want to set a date range for this so that it is only replaced pre 6/1/2020

grgmiller: Also, do we need to also update the total interchange value?

grgmiller (changes requested):

src/eia930.py
gailin-p: Good catch, total interchange does need to be changed. here's the raw data:
Screen Shot 2022-09-06 at 10 42 17 PM

gailin-p: Same series after fix. 

Screen Shot 2022-09-07 at 4 18 06 PM

src/eia930.py
```diff
@@ -441,12 +441,19 @@ def manual_930_adjust(raw: pd.DataFrame):
...
+    # Interchange AZPS - SRP is wonky before 6/1/2020 7:00 UTC. Use SRP - AZPS (inverted)
     azps_srp = get_int_columns("AZPS", raw.columns, ["SRP"])
     srp_azps = get_int_columns("SRP", raw.columns, ["AZPS"])
-    raw.loc[:, azps_srp] = (raw.loc[:, srp_azps] * (-1)).rename(
+    replacement = (raw.loc[:, srp_azps] * (-1)).rename(
         columns={srp_azps[0]: azps_srp[0]}  # rename so Pandas will do the right thing
     )
+    raw.loc[:"2020-06-01T07:00+00", azps_srp] = replacement[:"2020-06-01T07:00+00"]
+    # Update total interchange
+    all_cols = [c for c in get_int_columns("AZPS", raw.columns) if "ALL" not in c]
```
grgmiller, line 452: This should exclude the total interchange column from this list, correct? This will only return all of the BA to BA interchange columns?

gailin-p: yep, that's done by the `if "ALL" not in c` in the list comprehension. not the most readable but it is correct

grgmiller (approved):
Looks good!

Issue #231, opened 2022-09-15
Hourly data not validated for overlapping data


grgmiller:
It looks like in `data_cleaning.combine_plant_data()` a argument `validate` was added with a default value of `False`. When set to False, this function does not run `validation.ensure_non_overlapping_data_from_all_sources()`.

In the data pipeline, step 15, there is no validate argument specified when combining the hourly data, which means by default this data is not being checked for overlapping data.

This argument should always be set to `True` by default.

grgmiller:
It looks like this validation check was being run for the monthly data but not the hourly data because the validatation check compares data by subplant id, which is missing from the shaped EIA data. Since it is run for the monthly data export, there should be no duplication of data after that point.

Issue #232, opened 2022-09-15
Plants with multiple respondent frequencies


grgmiller:
Generally, plants report data to EIA-923 for all of their parts at a single frequency: monthly data or annual data. 

However, certain plants appear to report some of their data monthly (AM) and some data annually (A), such as plant 422:

![image](https://user-images.githubusercontent.com/45949268/190509935-93805b63-df10-4035-9c9b-c344f6dcfada.png)

We should check:
1. Is this correct that plants can report at different frequencies, or is this a typo?
2. If this multi-frequency reporting is correct, then is the way that the annual data is getting assigned to months correct? The way the EIA currently does this shaping is based on the monthly shape of generators with similar characteristics. However, if we have monthly data for part of the plant, would it be more realistic to assume that the part of the plant that reports annually follows a similar pattern to the plant that reports monthly?

Also, if it is the case that plants can have multiple reporting frequencies, we should figure out how to address this in pudl. Currently, reporting frequency is a plant-level attribute in the plants_eia860 table, but this might need to be updated.

PR #233, opened 2022-09-16
Add BA-level data quality metrics

grgmiller:
Previously, all of our data quality metrics only provided information about the quality of data for the entire country. However, the quality of data in individual BAs could vary quite widely. 

This PR is a work in progress, but will add BA-level metrics (in addition to the national metrics) to:
- [X] input_data_source
- [x] hourly_profile_method

By working on this, I also discovered that some plants report data at multiple frequencies (https://github.com/singularity-energy/open-grid-emissions/issues/232), which was causing some of the metrics to not sum to 100% because there was missing frequency codes. This patches that by filling missing frequency codes with "multiple" (as opposed to "annual" or "monthly")


grgmiller:
In addition to adding BA-level data quality metrics, I also added a notebook that can be used to explore stats for generators that do not report to CEMS and how they might differ. A lot of this work is based on feedback I got during some peer-review of my dissertation chapter.

grgmiller:
This PR also fixes several other bugs that were introduced in other PRs that have been merged into the v0.1.2 branch:
- At the beginning of the data pipeline, we were deleting the results folder for the year we were running, but this was raising an error if this directory didn't already exist (ie if this is the first time the user has ever run the pipeline). This fixes that by first looking to see if the directory exists already.
- In https://github.com/singularity-energy/open-grid-emissions/pull/245 I noticed that the validation check for overlapping data was not being run when combining hourly data, so I "fixed" this by setting the validate parameter to True. However, I discovered that there was actually a reason that this had been set to false (although it was not documented): The overlapping data validation checks the data at the subplant level, and when we combine hourly data, the `shaped_eia_data` df only contains fleet-level data and not subplant data so this validation check won't work. I set the validation parameter back to False when combining the hourly data and added a note about why we are doing that. (Closes https://github.com/singularity-energy/open-grid-emissions/issues/231)
- Fixes some SettingWithCopy warnings in `output_data.write_plant_metadata()`

gailin-p (approved):

Issue #234, opened 2022-09-16
Update pip dependencies


grgmiller:
The OGE pipeline depends on two external projects: `pudl` and `gridemissions`. However, our dependencies are currently to two forks that @gailin-p and I have made from the original code:
 - https://github.com/grgmiller/pudl.git@allocate_fuel
 - https://github.com/gailin-p/gridemissions

We should come up with a more permanent solution for this going forward. If we are going to continue working off of forks, we may want to fork these to the singularity-energy github account, instead of our personal acccounts.

With regards to the base repositories:
 - For PUDL: I think once the current updates in the dev branch of pudl (including the boiler fuel allocation improvements, once merged) are merged into main, we could point back to pudl@main. 
 - For gridemissions: I think our use of the base repo depends on how actively they want to maintain the base repository. If they are not interested in actively maintaining it going forward, we may want to adopt the repo and/or just start working off of a singularity fork. I don't think we necessarily want to create diverging versions of the same code, but we should work with them to figure that out.

gailin-p:
The context for this issue was changed by #235, which points at an archived version of PUDL code (hosted on @grgmiller's github) and data (hosted on Zenodo). Because these archived versions require us to manually update them and then check our own code for compatibility whenever we want to incorporate new PUDL data, they are not ideal. 

However, we probably don't want to point directly at pudl@main until PUDL releases a 1.x version, because PUDL in 0.x changes data formats and breaks our code, and it's nice to have control over when we bring ourselves up to date with these changes (ie, by archiving our own version of PUDL, #235)

grgmiller:
With v0.2.0, we now point directly to pudl@main for our pudl dependency.

grgmiller:
Fixed by https://github.com/singularity-energy/open-grid-emissions/pull/327 and https://github.com/singularity-energy/open-grid-emissions/pull/318

PR #235, opened 2022-09-22
update pudl dependency

grgmiller:
I created a new branch of my fork of pudl at https://github.com/grgmiller/pudl/tree/oge to use as the new dependency for the OGE repository. 

I had been using the same branch to contribute code to pudl that I had been using as the dependency for OGE. However, to make updates to pudl, I need to merge in updates from pudl's dev branch that would break our data pipeline.

This PR:
- updates the pudl dependency in the `environment.yml` file to use a stable fork of pudl
- Instead of downloading the updated `pudl.sqlite` file directly from catalyst's datasette (which is not stable), we now download the archived version of the file from our zenodo.

Once pudl releases a new version, we can update these dependencies.


grgmiller:
This advances https://github.com/singularity-energy/open-grid-emissions/issues/234

gailin-p (approved):

PR #236, opened 2022-09-23
Include net_consumed_mwh

miloknowles:


src/data_cleaning.py
```diff
@@ -4,7 +4,7 @@
 import sqlalchemy as sa
 
 import pudl.analysis.allocate_net_gen as allocate_gen_fuel
-import pudl.analysis.epa_crosswalk as epa_crosswalk
+import pudl.analysis.epacamd_eia as epa_crosswalk
```
grgmiller, line 7: This is a new change that pudl just released in their dev branch but that we're not ready to integrate yet. I just pushed an update to the main branch of OGE that updates the dependency in the environment so use the old code for now.

Issue #49, opened 2022-06-07
Ensure complete `subplant_id` mapping


grgmiller:
Currently, subplant IDs are only created for units that exist both in CEMS and EIA-923, meaning that there are certain generators/units that have a subplant ID of `NaN`.

- [x] Ensure that all merge and groupby functions that use `subplant_id` as one of the keys are not dropping observations with missing subplant values.
- [ ] Although the primary purpose of the subplant ID is to group CEMS units with EIA generators and boilers,  it could also be useful for grouping EIA boilers and generators that do not exist in CEMS. We should update the `pudl.analysis.epa_crosswalk` code to generate subplant IDs for all boilers/generators that exist in the EIA data, regardless of whether data exists in CEMS.
- [ ] If there are any remaining missing subplant values, we should perhaps fill these missing values with a code of 99 so that there is a non-missing code that would not overlap with any subplant ids already assigned during the crosswalk process.

gailin-p:
In some cases, generators in a single plant missing from subplant_crosswalk have a mix of renewable and fossil fuel types. This occurs in 74 subplant-months in plants 141, 621, 1943, 2240, 10025, 10823, and 58236. In these cases, all generators in the plant which are not in subplant_crosswalk are assigned the same subplant, `subplant_id=NaN`. 

In #230, we propsoed that subplants within a plant should not share the same CEMS profile (hourly shaping method `partial_cems_plant`) when they have different primary fuel types, since this resulted in one case where all nuclear generation from a large nuclear power plant `plant_id_eia=2410` was being assigned to the 3 hours where a backup diesel generator was on and reporting to CEMS. However, because renewable and fossil generators are combined in each of the subplants listed above, the renewable and fossil generators cannot be assigned different profiles. 

If the renewable and fossil generators were assigned different subplants, we could safely use `partial_cems_plant` to shape the subplant with the fossil generators and a residual profile method to shape the subplant with the renewable generators. This would be conceptually more correct than choosing one method to apply to a sublant with mixed fossil and renewable generation. 

To fix this, we would need to update subplant crosswalk (see @grgmiller 's comments above, we could potentially do this in PUDL) to assign different subplant IDs to generators within a plant whose fuel types differ. 



> adding subplant ids for EIA-only data might become irrelevant (at least for the initial public release) if we are grouping this data by BA-fuel anyway.

Since hourly data is shaped at the subplant level, I think this does end up affecting currently released data.

gailin-p:
[...]

Issue #50, opened 2022-06-07
Identify outlier values in reported CEMS data


grgmiller:
We should implement some sort of outlier detection and screening for the hourly values reported in CEMS. This outlier detection could use a combination of statistical methods and physics-based methods (e.g. gross generation should not exceed nameplate capacity).

This should probably be implemented after loading the CEMS data but before any missing data imputation steps.

grgmiller:
This paper includes some possible approaches to identifying outlier data: https://pubs.acs.org/doi/10.1021/acs.est.9b04522

PR #238, opened 2022-09-28
Gailin/issue230

### LINKED ISSUE #230: Spikes in emission rate data
Some BAs have negative spikes in emission rate data. (negative meaning they deviate negatively from the trend)

Eg, PJM: 
Screen Shot 2022-09-13 at 12 48 40 PM

WACM: 
Screen Shot 2022-09-13 at 12 50 11 PM

Both the above figures compare real-time (red) to OGE (blue). Both show adjusted emission rate (CO2/MWh). In the OGE data, the relevant variable is `generated_co2_rate_lb_per_mwh_for_electricity_adjusted` in `power_sector_data`. 

In some small BAs, rate spikes are expected (eg, DEAA where there is only one `natural_gas` plant and so plant startup leads to enormous positive BA-level emission rate spikes). However, we don't expect spikes that are abrupt *decreases* in rate (eg WACM), and we don't expect spikes in large BAs (eg, PJM)

gailin-p:
## Summary of changes

Close #230 by changing hourly data source identification methods: 
* A renewable generator (hydro, wind, solar, nuclear, geothermal) won't use the `partial_cems_plant` hourly data source methodology
* A subplant that contains generators of mixed fuel types will choose the hourly data source of the generator with the largest generation (this is relevant for subplants with `subplant_id=NaN`, which can have generators of mixed fuel types; see comments on #49)

The post-fix hourly emission rate for PJM is below. The abrupt drops to zero (seen in #230) are gone. 
Screen Shot 2022-09-28 at 9 22 30 AM

The PJM nuclear emission rate, below, shows the three hours where the generator is on as higher-than-zero emission rates: 
Screen Shot 2022-09-28 at 9 22 47 AM

## Open question: subplants in hourly `plant_data` results

Plant 2410, the nuclear plant originally responsible for the PJM data issues, has one subplant made up of a CEMS-reporting diesel generator and one subplant with its nuclear generation. Generation from the CEMS-reporting subplant is in `plant_data/hourly/individual_plant_data.csv`, while generation from the nuclear subplant is in `shaped_fleet_data.csv`. This is confusing, since a user looking only at `individual_plant_data.csv` would think that plant 2410 had only ~10 MWh of generation in 2020, but if they looked at the annual data, they'd see annual generation of 16,000,000 MWh. 

Proposed solutions: 
1. Exclude plants from `individual_plant_data.csv` if one or more of their subplants has `hourly_data_source=eia`
2. Make a note in the documentation that some plants in `individual_plant_data.csv` may not contain all subplants, and indicate how users can identify whether a plant has complete hourly data 

I think 1 is the better solution, but it does mean removing some hourly data that we're actually pretty confident in. (eg, for 2410, we do know that the diesel generator ran for those 3 hours).

### DIFF
--- a/src/data_cleaning.py
+++ b/src/data_cleaning.py
@@ -1476,9 +1476,10 @@ def identify_partial_cems_plants(all_data):
         "hourly_data_source",
     ] = "partial_cems_plant"
 
-    # It is possible to get plants with NaN-ID subplants with mixed hourly methods
-    # due to the hourly source code check above if subplant_id is mixing fuel types.
-    problem_subplant_months = (
+    # It is possible to have generators within a subplant assigned different hourly methods
+    # due to the source code check above if subplant_id is mixing fuel types.
+    # This was an issue before PR #239; this check will catch it if it reoccurs
+    mixed_method_subplants = (
         all_data.groupby(
             [
                 "plant_id_eia",
@@ -1490,15 +1491,15 @@ def identify_partial_cems_plants(all_data):
         .nunique()
         .reset_index()
     )
-    problem_subplant_months = problem_subplant_months[
-        problem_subplant_months.hourly_data_source > 1
+    mixed_method_subplants = mixed_method_subplants[
+        mixed_method_subplants.hourly_data_source > 1
     ]
-    if len(problem_subplant_months) > 0:
+    if len(mixed_method_subplants) > 0:
         # Check for subplants with mixed hourly data sources,
         # likely resulting from mixed fuel types.
         # If subplant_id assignment is working, there shouldn't be any
         raise Exception(
-            f"    ERROR: {len(problem_subplant_months)} subplant-months have multiple hourly methods assigned."
+            f"    ERROR: {len(mixed_method_subplants)} subplant-months have multiple hourly methods assigned."
         )
 
     # remove the intermediate indicator column
--- a/src/data_cleaning.py
+++ b/src/data_cleaning.py
@@ -1053,7 +1053,11 @@ def identify_hourly_data_source(eia923_allocated, cems, year):
             we will use the hourly values reported in CEMS. (Add a validation check for the net generation and fuel consumption totals)
         2. `partial_cems`: For subplant-months for which we have hourly CEMS data
             for only some of the CEMS units that make up a subplant, we will use the reported
-            EIA-923 values to scale the partial hourly CEMS data from the other units to match the total value for the entire subplant. This will also calculate a partial subplant scaling factor for each data column (e.g. net generation, fuel consumption) by comparing the total monthly CEMS data to the monthly EIA-923 data.
+            EIA-923 values to scale the partial hourly CEMS data from the other units to match the total value for the entire subplant.
+            This will also calculate a partial subplant scaling factor for each data column (e.g. net generation, fuel consumption) by
+            comparing the total monthly CEMS data to the monthly EIA-923 data.
+        2.5 `partial_cems_plant`: when some subplants in a plant have CEMS data, use the shape from those subplants to shape the remaining generation
+            from that plant.
         3. `eia`: for subplant-months for which no hourly data is reported in CEMS,
             we will attempt to use EIA-930 data to assign an hourly profile to the monthly EIA-923 data
     Inputs:
@@ -1272,15 +1276,73 @@ def identify_partial_cems_plants(all_data):
         validate="m:1",
     )
 
+    # Plants with these primary fuel types shouldn't be shaped using CEMS data --
+    # any cems data is likely to be a backup generator and therefore not representative of overall generation.
+    non_cems = ["WND", "WAT", "NUC", "SUN", "GEO"]
+
     # for plants where the data source is eia only, and it exists in partial plant, change the source to partial_cems_plant
+    # but only if the fuel type is consistent with CEMS fuel types
     all_data.loc[
         (all_data["hourly_data_source"] == "eia")
-        & (all_data["partial_plant"] == "both"),
+        & (all_data["partial_plant"] == "both")
+        & (
+            ~all_data["energy_source_code"].isin(non_cems)
+        ),  # CEMS data should only be used to shape if plant is primarilly fossil
         "hourly_data_source",
     ] = "partial_cems_plant"
 
+    # We might have some plants with NaN-ID subplants with mixed hourly methods
+    # due to the hourly source code check above. In those cases, we need to select one methodology -- otherwise we'll double-count the generation
+    # We'll choose the generator with the largest net_generation_mwh and use its methodology.
+    # If they're all zero, we'll choose the generator with the largest fuel_consumed_mmbtu
+    problem_subplant_months = (
+        all_data.groupby(["plant_id_eia", "subplant_id", "report_date"], dropna=False)
+        .hourly_data_source.nunique()
+        .reset_index()
+    )
+    problem_subplant_months = problem_subplant_months[
+        problem_subplant_months.hourly_data_source > 1
+    ]
+    print(
+        f"    WARNING: {len(problem_subplant_months)} subplant-months have multiple hourly methods assigned. Choosing one."
+    )
+
+    # for each problem subplant-month, identify the record with the largest net_generation_mwh or fuel_consumed_mmbtu
+    def subplant_month_to_method(rows):
+        rows.sort_values(by=["net_generation_mwh", "fuel_consumed_mmbtu"])
+        return rows.iloc[0].hourly_data_source
+
+    method_assignments = (
+        all_data[
+            all_data.plant_id_eia.isin(problem_subplant_months.plant_id_eia.unique())
+            & all_data.subplant_id.isna()
+        ]
+        .groupby(["plant_id_eia", "subplant_id", "report_date"], dropna=False)
+        .apply(subplant_month_to_method)
+    )
+    ma = method_assignments.reset_index().rename(
+        columns={0: "corrected_hourly_data_source"}
+    )
+    all_data = all_data.merge(
+        ma,
+        how="left",
+        on=["plant_id_eia", "subplant_id", "report_date"],
+        validate="m:m",
+    )  # multiple generators per subplant, so multiple records
+    to_replace = ~all_data.corrected_hourly_data_source.isna()
+    all_data.loc[to_replace, "hourly_data_source"] = all_data.loc[
+        to_replace, "corrected_hourly_data_source"
+    ]
+
     # remove the intermediate indicator column
-    all_data = all_data.drop(columns=["partial_plant", "eia_data", "cems_data"])
+    all_data = all_data.drop(
+        columns=[
+            "partial_plant",
+            "eia_data",
+            "cems_data",
+            "corrected_hourly_data_source",
+        ]
+    )
 
     return all_data
 
--- a/src/data_cleaning.py
+++ b/src/data_cleaning.py
@@ -1476,58 +1476,30 @@ def identify_partial_cems_plants(all_data):
         "hourly_data_source",
     ] = "partial_cems_plant"
 
-    # We might have some plants with NaN-ID subplants with mixed hourly methods
-    # due to the hourly source code check above. In those cases, we need to select one methodology -- otherwise we'll double-count the generation
-    # We'll choose the generator with the largest net_generation_mwh and use its methodology.
-    # If they're all zero, we'll choose the generator with the largest fuel_consumed_mmbtu
-    key_cols = [
-        "plant_id_eia",
-        "subplant_id",
-        "report_date",
-    ]  # not unique key, because same subplant may have multiple generators
+    # It is possible to get plants with NaN-ID subplants with mixed hourly methods
+    # due to the hourly source code check above if subplant_id is mixing fuel types.
     problem_subplant_months = (
-        all_data.groupby(key_cols, dropna=False)
-        .agg({"hourly_data_source": "nunique", "net_generation_mwh": "sum"})
+        all_data.groupby(
+            [
+                "plant_id_eia",
+                "subplant_id",
+                "report_date",
+            ],
+            dropna=False,
+        )
+        .nunique()
         .reset_index()
     )
     problem_subplant_months = problem_subplant_months[
         problem_subplant_months.hourly_data_source > 1
     ]
-
-    # May need to fix plants, may not (on small run we don't)
     if len(problem_subplant_months) > 0:
-        print(
-            f"    WARNING: {len(problem_subplant_months)} subplant-months totaling {problem_subplant_months.net_generation_mwh.sum()} MWh generation have multiple hourly methods assigned. Choosing one."
+        # Check for subplants with mixed hourly data sources,
+        # likely resulting from mixed fuel types.
+        # If subplant_id assignment is working, there shouldn't be any
+        raise Exception(
+            f"    ERROR: {len(problem_subplant_months)} subplant-months have multiple hourly methods assigned."
         )
-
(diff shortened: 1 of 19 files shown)


gailin-p:
Other changes: 

There are some notebook changes included here that are not directly related to #230: 
* `hourly_validation.ipynb` and `map_visualization.ipynb` are updated to reflect final changes I made before publishing the visualizations in the two blog posts (announcement and real-time validation posts)
* `work_in_progress/clean_cems_outliers.ipynb` is a notebook I pulled out of the old/outdated `gailin/clean_cems` branch. This notebook is background research to advance #50. Moving the notebook into `work_in_progress` will allow us to delete the outdated `gailin/clean_cems` branch.

grgmiller:
> Proposed solutions:
> 1. Exclude plants from individual_plant_data.csv if one or more of their subplants has hourly_data_source=eia
> 2. Make a note in the documentation that some plants in individual_plant_data.csv may not contain all subplants, and indicate how users can identify whether a plant has complete hourly data

So the original partial plant methodology was designed such that different parts of a plant couldn't end up in separate files: If all subplants had `hourly_data_source=="eia"` all of the plant data would end up in `shaped_fleet_data`, but if only subplants had eia as the data source, we would use one of the partial cems methods, and then all of the data for that plant would end up in `individual_plant_data` (or at least that was the intent - if you've found that's not how it was working then we would want to fix that bug).

However, with this current issue, we've discovered that we don't always want to shape an entire plant with partial cems data because of the spike issue. Currently, this PR now is splitting up the data so that (in the case of 2410 for example) the nuclear portion would end up in the shaped fleet data and the diesel portion would end up in the individual data. 

If we went with option 1, do you know how much CEMS data we'd be ignoring (ie is it a handful of backup generators with negligible generation, or would this affect a wider set of generation)?

I'm kind of leaning toward 2 as the simpler fix for now.

src/data_cleaning.py
```diff
@@ -1272,15 +1276,73 @@ def identify_partial_cems_plants(all_data):
...
+        & (all_data["partial_plant"] == "both")
+        & (
+            ~all_data["energy_source_code"].isin(non_cems)
+        ),  # CEMS data should only be used to shape if plant is primarilly fossil
         "hourly_data_source",
     ] = "partial_cems_plant"
 
+    # We might have some plants with NaN-ID subplants with mixed hourly methods
+    # due to the hourly source code check above. In those cases, we need to select one methodology -- otherwise we'll double-count the generation
+    # We'll choose the generator with the largest net_generation_mwh and use its methodology.
```
grgmiller, line 1296: Can you provide some summary information about how much generation this method impacts? Does this happen to affect more plants than just the specific situation of backup generators that run more than a handful of hours per year? How common is NA subplants with mixed fuels?

grgmiller (changes requested):

src/data_cleaning.py
```diff
@@ -1272,15 +1276,73 @@ def identify_partial_cems_plants(all_data):
...
+    non_cems = ["WND", "WAT", "NUC", "SUN", "GEO"]
+
     # for plants where the data source is eia only, and it exists in partial plant, change the source to partial_cems_plant
+    # but only if the fuel type is consistent with CEMS fuel types
     all_data.loc[
         (all_data["hourly_data_source"] == "eia")
-        & (all_data["partial_plant"] == "both"),
+        & (all_data["partial_plant"] == "both")
+        & (
+            ~all_data["energy_source_code"].isin(non_cems)
```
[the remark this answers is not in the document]
gailin-p: `all_data` is at the generator level, not the subplant level. This addition does potentially break a subplant into multiple methodologies if some generators in a subplant are renewable and some aren't; the rest of this method checks for those cases and assigns all generators in subplants the methodology used by the largest generator.

src/data_cleaning.py
gailin-p: This occurs in 74 subplant-months in plants 141, 621, 1943, 2240, 10025, 10823, and 58236. It is not always a case of backup generators; for example, 141 is mostly a natural gas plant with one small solar array with zero generation reported to EIA-923.

gailin-p: There is a total of 13034375 MWh over the affected subplant-months

src/data_cleaning.py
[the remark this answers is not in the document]
grgmiller: Ok makes sense. How has this approach been working for the handful of generators this affects?

src/data_cleaning.py
```diff
@@ -1272,15 +1276,73 @@ def identify_partial_cems_plants(all_data):
         validate="m:1",
     )
 
+    # Plants with these primary fuel types shouldn't be shaped using CEMS data --
+    # any cems data is likely to be a backup generator and therefore not representative of overall generation.
+    non_cems = ["WND", "WAT", "NUC", "SUN", "GEO"]
```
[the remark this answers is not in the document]
gailin-p: Yes, I confirmed that there are no generators with `plant_primary_fuel==GEO` that have CEMS data

gailin-p:
>If we went with option 1, do you know how much CEMS data we'd be ignoring (ie is it a handful of backup generators with negligible generation, or would this affect a wider set of generation)?

The new methodology proposed in this PR results in 106 plant-months (over 12 unique plants) with split EIA and CEMS/partial_CEMS hourly data sources. The generation from these plants is split pretty evenly over CEMS and EIA hourly data sources, with about 21,000,000 MWh generation in CEMS subplants and 29,000,000 in EIA subplants. 

The affected plants are: [557, 621, 645, 1355, 2410, 2707, 2953, 6074, 8223, 10029, 10823, 58236]. (clarification note: these are plants with split methodologies *between* subplants, which is a different set of plants than the plants with NaN-id subplants with split methodologies *within* a subplant.)

gailin-p:
[725e39a](https://github.com/singularity-energy/open-grid-emissions/pull/238/commits/725e39a76c234d9876d491817160d58968a5afb5) added a new hourly plant-level output file, partial_plant_data.csv, which contains the CEMS and partial CEMS plant data for plants with one or more subplants without hourly data.

src/data_pipeline.py
```diff
@@ -176,6 +177,14 @@ def main():
     eia923_allocated = data_cleaning.identify_hourly_data_source(
         eia923_allocated, cems, year
     )
+    # Export data cleaned by above for later validation, visualization, analysis
```
grgmiller, line 180: Why did you move this export up in the code? The original placement was meant to export it after the hourly data source column had been updated by the partial CEMS shaping process

grgmiller (changes requested):

grgmiller:
The more I think about it, I'm thinking that perhaps we should revert https://github.com/singularity-energy/open-grid-emissions/pull/238/commits/725e39a76c234d9876d491817160d58968a5afb5 and just keep the outputs split among the two files with a disclaimer in the documentation (and maybe even on the download page) that data might be split between two files. Here's my thought process:
-  I think that maybe in the next release, we will actually want to move to publishing all of the hourly plant-level data for individual plants, instead of being split into two files. In talking to some of the early adopters of the OGE dataset, it seems that there is interested in the individual plant data at the hourly resolution. In the past, we haven't done this partially due to memory issues, but I think there may be a way around this: if we shape the individual plant data in chunks only during export, and do not actually store a full version of the shaped plant-level data in memory, I think this could work. We'd want to export the data in chunks (eg for each BA, or each state) so that each file isn't huge. Once we export the hourly plant level data, we could apply the existing aggregated shaping method since that's all we need for the subsequent power sector and consumed outputs. 
- For a patch release, I'm not sure that we want to be adding a whole new output to the dataset, especially if it is a temporary fix that we will probably not keep in the next major release. 
- It seems like the fix for outputting data into a separate file involves re-organizing the data pipeline a bit, and I'd want to review this a bit more carefully before implementing.

gailin-p:
>It seems like the fix for outputting data into a separate file involves re-organizing the data pipeline a bit, and I'd want to review this a bit more carefully before implementing. 

Most of the reorganization is actually unrelated -- I just moved the writing of eia923 to just after the data is finished because I found it confusing for it to be exported later in the pipeline when it hasn't actually been modified since line 172. The only other pipeline change is to not delete it so it can be passed to the plant writing function. 

>For a patch release, I'm not sure that we want to be adding a whole new output to the dataset, especially if it is a temporary fix that we will probably not keep in the next major release.

I see your point here, though it doesn't feel too problematic to me (it'll still get zipped up with the rest of the hourly plant data). 

I think if we want to bundle the partial hourly plant data with the regular plant data, we should write a new data_quality_metrics file that reports the plant-months for which hourly plant data is partial. There's no other way to get that info without the `outputs` files. (We should maybe also copy that file to the `plant_data/hourly` folders so users get that information without having to search out and download another zip)

src/data_pipeline.py
grgmiller: Oh never mind! The old placement was leftover from a previous iteration where the partial CEMS shaping modified this dataframe, but that's no longer the case so this move definitely makes sense!

grgmiller:
> I found it confusing for it to be exported later in the pipeline when it hasn't actually been modified since line 172

Oh sorry about that - I realize now that in a previous iteration of the code the `eia923_allocated` df was modified by the partial CEM shaping process, but that is no longer the case, so your move makes total sense!

> we should write a new data_quality_metrics file that reports the plant-months for which hourly plant data is partial

Doesn't `results/2020/plant_data/plant_metadata.csv` already do this? For each subplant-month, it identifies the source of the data and the source of the hourly profile. Without looking at a specific plant example, I'm not sure if this output needs to be further modified to meet these needs, but we should be able to use this output file.

src/data_cleaning.py
```diff
@@ -1457,15 +1461,54 @@ def identify_partial_cems_plants(all_data):
...
+        & (all_data["partial_plant"] == "both")
+        & (
+            ~all_data["energy_source_code"].isin(non_cems)
+        ),  # CEMS data should only be used to shape if plant is primarilly fossil
         "hourly_data_source",
     ] = "partial_cems_plant"
 
+    # It is possible to get plants with NaN-ID subplants with mixed hourly methods
+    # due to the hourly source code check above if subplant_id is mixing fuel types.
+    problem_subplant_months = (
```
grgmiller, line 1481: For clarity, could this be changed to something like `subplants_with_mixed_hourly_methods`

src/data_cleaning.py
```diff
@@ -1457,15 +1467,57 @@ def identify_partial_cems_plants(all_data):
...
+            ~all_data["energy_source_code"].isin(non_cems_fuels)
+        ),  # CEMS data should only be used to shape if plant is primarilly fossil
         "hourly_data_source",
     ] = "partial_cems_plant"
 
+    # Because we exclude clean generators from the partial_cems_plant method, it
+    # is possible that data for a single plant will end up in two different 
+    # output files. Thus, we want to identify if there are any plants that have
+    # a mix of methods.
+    problem_subplant_months = (
```
grgmiller, line 1490: I just want to confirm: are we trying to prevent mixed methods within a single subplant, or mixed methods within a single plant? This check is not trying to prevent, for example, a nuclear subplant and a diesel subplant from ending up in two different files, only the case where for some reason a nuclear generator and diesel generator were lumped into the same subplant?

grgmiller: I also think we should maybe update this variable name to be slightly more descriptive, eg `subplant_month_with_mixed_hourly_methods`

src/output_data.py
```diff
@@ -134,31 +134,63 @@ def output_data_quality_metrics(df, file_name, path_prefix, skip_outputs):
...
+        of its data is written or modified here (optional, only needed for hourly)
 
     Note: plant-level does not include rates, so all aggregation is summation.
     """
     if not skip_outputs:
         if resolution == "hourly":
             # output hourly data
             # Separately save real and aggregate plants
+            # Separately save complete plant CEMS and subplant CEMS
+            partial_plants = _identify_mixed_method_plants(eia923)
```
grgmiller, line 175: Is this still necessary if we are just using `results/plant_data/plant_metadata.csv` to indicate when a plant is split up?

grgmiller (changes requested):

gailin-p:
Ok, I like the idea of re-combining the hourly plant-level output files and directing users to plant_metadata to identify which plants have hourly data split between the plant file and the synthetic plant file. 

> Doesn't results/2020/plant_data/plant_metadata.csv already do this? For each subplant-month, it identifies the source of the data and the source of the hourly profile. Without looking at a specific plant example, I'm not sure if this output needs to be further modified to meet these needs, but we should be able to use this output file.

Great point! This is definitely the right spot for this data. However, it will need to be modified, because it doesn't currently contain rows for EIA-only plants (only rows for the synthetic plants aggregated from many EIA plants). For example, using the classic example plant for this issue, 2410 (the PJM nuclear plant with the diesel backup): we see metadata rows for the three months when the diesel subplant reported to CEMS, but no `plant_metadata.csv` row for the nuclear subplant -- it's just grouped in the synthetic plant row. 

Two options for fixing: 
1. Add rows only for shaped subplants of plants with some individual plant data (ie, the plant-months currently getting written to `partial_plant_data.csv` 
2. Add rows for all subplant-months. This would make for a much larger file, but it guarantees that a user will be able to look up where the hourly data is for any plant they're interested in

src/data_cleaning.py
gailin-p: this check prevents mixed methods within a single subplant. It's functionally a validation check for the mixed NaN-id subplants that were fixed by your subplant_id PR

gailin-p: clarified comment and var name in [84e4a9e](https://github.com/singularity-energy/open-grid-emissions/pull/238/commits/84e4a9ebf9d56a0f69684ac876a54c3cbb11880b)

grgmiller:
I think either option 1 or 2 would work, and we should do whatever is easiest to implement at this point (based on my educated guess, 2 might be easier since it involves less segmenting of the data, but I could be mistaken). The `plant_static_attributes` file also contains the mapping of plant id to shaped plant id, so in the worst case, a user would have to cross reference the two tables to figure out the metadata for a plant that had been aggregated.

src/output_data.py
```diff
@@ -300,15 +307,23 @@ def write_plant_metadata(
...
         cems["data_source"] = "CEMS"
         partial_cems_subplant["data_source"] = "EIA"
         partial_cems_plant["data_source"] = "EIA"
         shaped_eia_data["data_source"] = "EIA"
+        monthly_eia_to_shape["data_source"] = "EIA"
 
         # identify net generation method
         cems = cems.rename(columns={"gtn_method": "net_generation_method"})
         shaped_eia_data["net_generation_method"] = shaped_eia_data["profile_method"]
+        monthly_eia_to_shape["net_generation_method"] = "synthetic_plant"
```
grgmiller, line 326: So my understanding is that the plant metadata file will now have metadata for both the synthetic plant and all of the individual plants that make up a synthetic plant? If so, I think there are two options here: 
1. we might want to change this to "see shaped plant ID" so that it is clear that this is not a method, but instead telling the user to reference the methods for the synthetic plant id listed in the `shaped_plant_id` column
2. We could instead just copy the method used for the associated shaped plant id here

grgmiller: If we go with option 2, we ptobably need to merge in the `shaped_plant_id` column first so we can use it as a key

grgmiller (changes requested):

src/output_data.py
gailin-p: Went with option 1

gailin-p:
[07853b7](https://github.com/singularity-energy/open-grid-emissions/pull/238/commits/07853b702968d8ee66f4678d1a3e8ea75a023e4a) 
Removed separated full and partial plant-level hourly outputs 
Renamed metadata id to "<See shaped plant ID">

grgmiller (approved):
Looks good! I think we're ready to merge

PR #239, opened 2022-10-01
Update subplant mapping

### LINKED ISSUE #49: Ensure complete `subplant_id` mapping

grgmiller:
This PR closes #49 
This PR advances #230 

### Background
The primary purpose of this PR is to fix several issues related to `subplant_ids`:
- Ensures that there is a non-NA `subplant_id` mapped to every `generator_id` or `unitid`
- Ensures that the `subplant_id` actually takes known boiler-generator associations into account, as intended

These two general issues were leading to specific downstream issues in the data pipeline:
- Whenever generators or units had a `subplant_id` of NA, they were treated as a single subplant which was leading to inaccurate hourly shaping of data using the partial_cems methods when these generators had different fuel types (see https://github.com/singularity-energy/open-grid-emissions/issues/49#issuecomment-1259852035)
- We had previously thought that the pudl `epa_crosswalk` module was taking the EIA-860 boiler-generator association table into account when assigning subplant_ids, such that each subplant_id represented an interconnected network of units, generators, and boilers. However, inspecting the code more closely revealed that it actually ignores boiler-generator associations, so the subplant_id only represents clusters of units to generators. This meant that previously, the code was sometimes assigning separate subplant_id to generators that were connected to a single boiler (eg in the case of a combined cycle plant). This could lead to innacurate results when calculating gross to net ratios, or when aggregating the data later in the pipeline.

### How does this PR fix these issues?
One way to approach fixing this issue would be to actually update the pudl analysis module that creates the subplant_id based on graph analysis. At this time, doing so is not possible until the next versioned release of pudl is available. Instead, we perform some post-processing steps in our code that expand and correct the assigned subplant_ids.

1. **Ensure that all `generator_id` and `unitid` are included in the crosswalk table**. Previously, the subplant crosswalk table only included units/generators for which there was an identified match in the EPA power sector crosswalk table. However, there are plenty of plants that don't report to CEMS that could have a subplant_id assigned. We now include observations for all units that report to CEMS and all generators that report to EIA.
2. **Use boiler-generator associations to complete/update subplant_id mappings**. Similar to the graph analysis pudl uses to identify subplant_id, pudl also uses graph analysis to identify clusters of boilers and generators, to which it assigns a `unit_id_pudl`. We merge these `unit_id_pudl` into the subplant crosswalk and use these to update the subplant_id

This is the description of what the `update_subplant_id()` function does:
```
Data Preparation
Because the existing subplant_id crosswalk was only meant to map CAMD units to EIA generators, it
is missing a large number of subplant_ids for generators that do not report to CEMS. Before applying this
function to the subplant crosswalk, the crosswalk must be completed with all generators by outer
merging in the complete list of generators from EIA-860 (specifically the gens_eia860 table from pudl).
This dataframe also contains the complete list of `unit_id_pudl` mappings that will be necessary.

This function follows several steps:
1. Because the current subplant_id code does not take boiler-generator associations into account,
        there may be instances where the code assigns generators to different subplants when in fact, according
        to the boiler-generator association table, these generators are grouped into a single unit based on their
        boiler associations. The first step of this function is thus to identify if multiple subplant_id have
        been assigned to a single unit_id_pudl. If so, we replace the existing subplant_ids with a single subplant_id.
        For example, if a generator A was assigned subplant_id 0 and generator B was assigned subplant_id 1, but
        both generators A and B are part of unit_id_pudl 1, we would re-assign the subplant_id to both generators to
        0 (we always use the lowest number subplant_id in each unit_id_pudl group). This may result in some subplant_id
        being skipped, but this is okay because we will later renumber all subplant ids (i.e. if there were also a
        generator C with subplant_id 2, there would no be no subplant_id 1 at the plant).
        Likewise, sometimes multiple unit_id_pudl are connected to a single subplant_id, so we also correct the 
        unit_id_pudl basedon these connections.
2. The second issue is that there are many NA subplant_id that we should fill. To do this, we first look at
        unit_id_pudl. If a group of generators are assigned a unit_id_pudl but have NA subplant_ids, we assign a single
        new subplant_id to this group of generators. If there are still generators at a plant that have both NA subplant_id
        and NA unit_id_pudl, we for now assume that each of these generators constitutes its own subplant. We thus assign a unique
        subplant_id to each generator that is unique from any existing subplant_id already at the plant.
       In the case that there are multiple unitid at a plant that are not matched to any other identifiers (generator_id, 
       unit_id_pudl, or subplant_id), as is the case when there are units that report to CEMS but which do not exist in the EIA
       data, we assign these units to a single subplant.
```

### Other updates
- Improves the speed of running the subplant_id process by only loading the data that we needed. The function loads 5 years of CEMS data to identify which CEMS ids actually exist in the data that we might be using, but to do so, we were loading all of the gross generation data from CEMS using `load_data.load_cems_gross_generation`. This behavior was leftover from when the subplant_id and gross to net generation factor processes were combined. We created a new function `load_data.load_cems_ids` that only loads the id columns. Implementing this new function makes this data loading process 4x faster than the previous process.
- The subplant id function also included code to add proposed operating dates and retirement dates for each generator to the crosswalk. This code was split into a separate function for readability.
- Updates the name of `outputs/subplant_crosswalk.csv` to `outputs/subplant_crosswalk_{year}.csv` since the subplant_id are not stable across years. 
- Adds the adjusted R squared values for the gross to net regressions to the `outputs/gross_to_net_ conversions` file per the request of some users who were interested in exploring these values
- Adds several new manual unit to generator crosswalks to the `manual/epa_eia_crosswalk_manual` file
- Consolidates the global variable `CLEAN_FUELS` to a single location in `src/emissions`. Previously, this variable had been duplicated in the emissions, data_cleaning, and validation modules

# TODO
- [x] delete the notebook before merging


grgmiller:
I also added a notebook to start exploring issue #240

src/data_cleaning.py
```diff
@@ -142,8 +125,195 @@ def generate_subplant_ids(start_year, end_year, cems_monthly):
...
+        will only properly work when applied to a single plant_id_eia at a time.
+
+    Data Preparation
+        Because the existing subplant_id crosswalk was only meant to map CAMD units to EIA generators, it
+        is missing a large number of subplant_ids for generators that do not report to CEMS. Before applying this
+        function to the subplant crosswalk, the crosswalk must be completed with all generators by outer
+        merging in the complete list of generators from EIA-860 (specifically the gens_eia860 table from pudl).
+        This dataframe also contains the complete list of `unit_id_pudl` mappings that will be necessary.
+
+    This function follows several steps:
```
gailin-p, line 208: Could you either add a high-level summary of the hierarchy of subplant ID assignments? My understanding is that this hierarchy is: 
1) use the PUDL subplant_id. (in the case where the PUDL unit_id groups several subplants, use the PUDL unit_id) 
2) where there is no PUDL subplant_id, use the PUDL unit_id as the subplant_id
3) where there is neither a pudl subplant_id nor unit_id, use the generator ID as the subplant_id

src/data_pipeline.py
```diff
@@ -129,7 +129,7 @@ def main():
     ####################################################################################
     print("2. Identifying subplant IDs")
     # GTN ratios are saved for reloading, as this is computationally intensive
-    if not os.path.exists(outputs_folder(f"{year}/subplant_crosswalk.csv")):
+    if not os.path.exists(outputs_folder(f"{year}/subplant_crosswalk_{year}.csv")):
```
gailin-p, line 132: Since this is faster now, should we re-run with every pipeline run?

gailin-p (changes requested):

notebooks/work_in_progress/GH240_eia930_physics_reconciliation.ipynb
```diff
@@ -0,0 +1,138 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This notebook is set up to test code as needed."
```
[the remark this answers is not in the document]
grgmiller: That's a good question. In general, I was thinking that any notebooks that are actual unresolved works in progress we might want to keep in the repo until we've actually closed the associated issue. Once closed, I think we want to delete the notebook. 

There is content in this notebook that would be helpful (IMO) when we revisit this issue. I can add a note to the top of the notebook though to help us remember when we go back to this.

src/data_cleaning.py
```diff
@@ -142,8 +125,195 @@ def generate_subplant_ids(start_year, end_year, cems_monthly):
...
+    # We will use unit_id_pudl to fill missing subplant ids, so first we need to fill any missing unit_id_pudl
+    # We do this by assigning a new unit_id_pudl to each generator that isn't already grouped into a unit
+
+    # create a numeric version of each generator_id
+    # ngroup() creates a unique number for each element in the group
+    subplant_crosswalk["numeric_generator_id"] = subplant_crosswalk.groupby(
+        ["plant_id_eia", "generator_id"], dropna=False
+    ).ngroup()
+    # when filling in missing unit_id_pudl, we don't want these numeric_generator_id to overlap existing unit_id
+    # to ensure this, we will add 1000 to each of these numeric generator ids to ensure they are unique
```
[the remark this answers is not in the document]
grgmiller: My understanding is that `unit_id_pudl` is an integer ID assigned by the pudl pipeline for parts of each plant that always starts at 1, so in order to have an id > 1000, a single plant would have to have more than 1000 boilers/generators. Currently, the largest `unit_id_pudl` for a single plant is "10". 1000 was chosen as an arbitrarily high number.

src/data_pipeline.py
grgmiller: Good point, probably would be good practice.

src/data_cleaning.py
grgmiller: Nice, I like this summary! I'll add something like this.

gailin-p (approved):

Issue #241, opened 2022-10-02
Export hourly individual plant data for shaped plants


grgmiller:
Currently our hourly plant-level results are aggregated into two different files: the hourly data directly from CEMS is reported at the plant level, but the shaped EIA data is aggregated to the fleet data. Part of the reasoning for this was because there are simply a large number of shaped EIA plants, so attempting to shape each individual plant was causing memory errors (see https://github.com/singularity-energy/open-grid-emissions/issues/86#issuecomment-1153240982).

However, in speaking with some data users, it hourly plant level data for all individual plants would actually be useful, as the fleet-level aggregations are not useful for certain research. In addition, splitting the data into separate files based on the method used can cause issues if different methods are used for data from a single plant - hourly data for a single plant is in two different places so would need to be aggregated by the user to be complete (see https://github.com/singularity-energy/open-grid-emissions/pull/238). 

One potential way around the memory issue would be if we shape the individual plant data in chunks only during export, and do not actually store a full version of the shaped plant-level data in memory. We'd want to export the data in chunks (eg for each BA, or each state) so that each file isn't huge. Once we export the hourly plant level data, we could apply the existing aggregated shaping method since that's all we need for the subsequent power sector and consumed outputs.

PR #246, opened 2022-10-19
Add hourly data for all individual plants

grgmiller:
This PR is meant to address https://github.com/singularity-energy/open-grid-emissions/issues/241

NOTE: I'm requesting to merge into the pudl_update branch, not directly into development.

Updates in this PR: 
- Adds code to export hourly shaped data for each individual plant, instead of aggregating the eia_only data to the fleet level. We get around the memory issue by shaping and exporting plant data for each region (BA or state), so that we are not trying to hold hourly data for each plant for the entire country in memory at once. We add a new step 14 in the pipeline that shapes and exports this data, but does not retain this shaped data for the rest of the pipeline. Instead, after this export, the pipeline continues as it did previously, aggregating the data to the fleet level before shaping. This shaped fleet level data is used for the rest of the pipeline (calculating and exporting power sector and consumed emissions data) since individual plant data is not necessary for these steps. 
- I wrote in a new command-line argument `--shape_individual_plants` that gives the user the option to export individual plant data, or to skip that step, and export the aggregated fleet data as before. Shaping individual plant data will be the new default behavior.
- When exporting individual plant data, the pipeline only shapes and exports data for one subset of plants at a time, based on the location of the plant. The default behavior is to split up the data by balancing authority. Since BAs range in size, this still means that the biggest individual files are still pretty large (MISO is 1.4 GB), while some of the smaller BAs are less than 1MB. An alternative approach to more evenly distributing file size could be to export the data by state, although this could make it more difficult to work with the data on the other end if you want data for all of the plants in a certain BA (the function includes an option to use state as the export grouping, but ba_code is the default). 
- The output files from this process still takes up a large amount of space (about 8GB total), so to save memory, I had to limit the amount of data we are exporting:
     - The default behavior is to only export data in us_units and not metric. Exporting both would double the amount of disk space from these outputs (from 8 to 16 GB). The thought here is that researchers or others using this type of data can do the conversions themselves easily. 
     - I also limited the types of emissions columns that were being exported. We have 5 different pollutants and four ways of potentially presenting these data (raw, for electricity, adjusted, and for electricity adjusted), which would create 20 columns, and significantly increase the size of the files. For this output, I limited it to 11 emissions columns: the raw emissions data for each pollutant, the electricity-only emissions for each pollutant, and the adjusted electricity-only CO2 emissions. These seem to be the most relevant outputs for plant-level data uses, and limits duplicate data (eg CH4 doesn't actually get adjusted for biomass, so the ch4_mass_adjusted = ch4_mass). 
- Cleanup: Remove the `--gtn_years` argument since it is no longer relevant after improving the gross to net calculations and separating the gross to net calcuations from the subplant id identification process. The `identify_subplants()` function retains teh `number_of_years` parameter, but now this is just specified as 5 by default.


src/data_pipeline.py
```diff
@@ -375,9 +374,38 @@ def main():
...
+    hourly_profiles = impute_hourly_profiles.convert_profile_to_percent(
+        hourly_profiles,
+        group_keys=["ba_code", "fuel_category", "profile_method"],
+        columns_to_convert=["profile", "flat_profile"],
+    )
+
+    # 14. Export hourly plant-level data
+    ####################################################################################
+    print("14. Exporting Hourly Plant-level data for each BA")
+    if args.shape_individual_plants:
```
gailin-p, line 386: Should we skip this if --small?

gailin-p (approved):

src/data_pipeline.py
grgmiller: Good idea, I can add that.

Issue #255, opened 2022-11-09
Missing emissions control IDs leading to overestimation of NOX emissions


grgmiller:
When calculating controlled NOx and SO2 emissions, we rely on NOx emission rate and SO2 removal efficiency data from EIA-923 Schedule 8C ("Air Emissions Control Info"). We had been assuming that if a NOx emission rate was reported, that a corresponding NOx Control ID would also be reported, so when matching these emission rates to boilers, we relied on the NOx control ID as one of the primary keys.

However, examining our NOx results revealed they were higher than both the eGRID and EIA estimates by about 6-7%. It appears that the origin of this overestimation is that emissions control information for certain plants was not being merged because it is missing a NOx control ID, so for these plants we were calculating their emissions as the uncontrolled value.

One specific example is plant ID 10118 where we were calculating 1.4 million lb of NOx, compared to about 500 lb reported in eGRID. For this plant, the controlled NOx rates were associated with a mercury control ID, but not a NOx control ID. There are other examples where the NOx rate data is missing a NOX control ID, but is associated wth a PM control ID, SO2 control ID, or Mercury Control ID.

To fix this, we may need to load all of the different control IDs and use the other control IDs for other pollutants if the NOx control ID is missing. I am assuming that the order in which we check non-missing IDs shouldn't affect which boiler the data is associated with (ie whether we check the Hg ID first, then PM, then SO2, etc), but we should probably test this hypothesis by performing the boiler association using all of the IDs and making sure they are all the same.

This issue also affects the SO2 data, and when we add mercury or PM in the future, it would also affect those.

This also raises a question for EIA: why is there not a NOx control ID associated with the NOx emission rate data?

grgmiller:
So after fixing this issue, it turns out that this was only affecting a small portion of the NOx emissions - our results are still significantly higher than both the eGRID and EIA data. It appears that maybe a significant portion of the difference with eGRID results from missing fuel consumption data in eGRID for certain plants. However, this does not explain why we are still much higher than the EIA estimates. This should be the next area we look into.

PR #256, opened 2022-11-12
Fix missing NOx and SO2 identifiers

grgmiller:
This PR is meant to address https://github.com/singularity-energy/open-grid-emissions/issues/255.

To get around the fact that some NOx and SO2 control data doesn't have an associated nox or so2 control id, we now load all four control ids (nox, so2, pm, hg) and use these to map the data to a boiler if the nox or so2 id is missing. 

Previously we had dropped all rows from the air emissions control table that had a missing pollutant-specific control id, but now we only drop rows that don't have any pollutant-specific data (ie if we're calculating nox, we drop any rows that don't have any nox control data.

I've set this up as an iterative process where the user specifies which id they want to use as primary, and then the code cycles through each additional id, filling missing boiler associations until all data is associated with boilers (or there are no remaining ids to use). I did it this way instead of loading all control-boiler associations because many control ids are associated with many boilers, so this led to duplicate boiler matches. This is all coordinated by `emissions.associate_control_ids_with_boiler_id()`

Because we now need to load four different control id to boiler association tables, I replaced the nox- and so2-specific functions to load these tables with a more generic function `load_boiler_control_id_association_eia860()` in which the user can specify which pollutant table they want to load (eg nox, so2, hg, pm). This is possible because all four of these tables are in the same format.

### DIFF
--- a/src/emissions.py
+++ b/src/emissions.py
@@ -975,26 +975,28 @@ def load_controlled_nox_emission_rates(year):
     # load the emissions control data
     emissions_controls_eia923 = load_data.load_emissions_controls_eia923(year)
 
-    # create a dataframe that contains only operating NOx emission control data
-    nox_rates = emissions_controls_eia923[
-        ~emissions_controls_eia923["nox_control_id"].isna()
-    ]
+    # create a dataframe that contains only NOx emission data for operating control equipment
+    nox_rates = emissions_controls_eia923.dropna(
+        axis="index",
+        how="all",
+        subset=[
+            "annual_nox_emission_rate_lb_per_mmbtu",
+            "ozone_season_nox_emission_rate_lb_per_mmbtu",
+        ],
+    )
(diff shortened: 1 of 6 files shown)


src/emissions.py
```diff
@@ -1063,6 +1052,76 @@ def calculate_weighted_nox_rates(year, nox_rates, aggregation_level):
...
 
+def associate_control_ids_with_boiler_id(df, year, pollutant):
+    """Associates emission control ids with boiler_id.
+
+    Because some emission control data is missing a corresponding control id for that pollutant,
+    this function attempts to use all control ids to associate each obervation with a boiler id,
+    in the order specified by `id_order`
+    """
+
+    all_pollutants = ["pm", "so2", "nox", "hg"]
```
gailin-p, line 1063: I assume that changing this order causes few or no differences in the resulting map between emission control IDs and boiler_id?

grgmiller: That is correct - that is something I checked.

Issue #258, opened 2022-11-19
Update OGE to work with the next PUDL release / 2021 data


grgmiller:
The latest release of PUDL includes [many updates](https://catalystcoop-pudl.readthedocs.io/en/latest/release_notes.html) and potentially breaking changes that we will need to address. The changes I've flagged so far include:

- [x] update the environment to be compatible with pudl
- [x] Update code/warnings to allow 2021 data
- [x] The EPA-EIA crosswalk is now integrated into pudl. We should look into whether we still need to separately download this
- [x] `prime_mover_code` has moved from `generators_entity_eia` to `generators_eia860` since it is no longer considered a static attribute
- [x] Check for updates to the `operational_status_code` encoding
- [x] Several attributes were moved from `plants_entity_eia` to `plants_eia860`, including `balancing_authority_code_eia`, `balancing_authority_name_eia`, grid voltage columns, and iso codes
- [x] `grid_voltage_kv` was renamed to `grid_voltage_1_kv`
- [x] Check `balancing_authorities_eia` for changes to BA encoding, including changes to PACW and PACE
- [x] Several columns in CEMS were renamed: `unitid` -> `emissions_unit_id_epa`. `facility_id` was dropped.
- [x] Missing values in CEMS are no longer replaced with zeros - no longer need method to re-create these missing values.
- [x] May no longer need to convert `plant_id_eia` in CEMS to `plant_id_epa` before converting to `plant_id_eia`
- [x] There were changes made to `allocate_net_gen` before mergining - we should double check that everything still works.

grgmiller:
Additional to dos:
- [ ] Consider deleting `load_data.crosswalk_epa_eia_plant_ids()` if it is not used by any functions anymore. pudl is doing this crosswalk but not including all manual CW.
- [x] Double check that we don't have any dependencies on eGRID since the 2021 version is not yet published
- [ ] Double check all of the manual cleaning of EIA-930 data is updated for year 2021 data
- [ ] Consider adding steam-only units back to OGE if https://github.com/singularity-energy/open-grid-emissions/issues/216 is correct
- [ ] Go through all files in `data/manual` to update for 2021 if necessary
- [ ] Test that co2-eq functions are using AR6 for 2021 data

Meta to-do:
- [ ] Create a checklist of everything we need to check / update each time a new year of data is released
- [ ] Consider whether it is worth running the PUDL pipeline locally in the future instead of waiting for PUDL release.

[...]

Issue #163, opened 2022-08-02
Using fuel-weighted factors for filling all missing CO2 data in CEMS


grgmiller:
Currently, when filling missing co2 data in cems (using `emissions.fill_cems_missing_co2`), we first calculate GHG emissions from fuel consumption based on the single fuel type reported for that unit in the Power Sector Data Crosswalk. Then, for units for which a single `energy_source_code` is not identified, we calculate an emission factor for that unit based on the month-specific, weighted average emission factor of the fuel consumed by that unit as reported in EIA-923.

It seems like the second method may actually be more accurate/specific, and the one that we should primarily use. For units that may burn multiple fuels in a month, or which fuel switch throughout the year, this would give us a more accurate calculation of emissions than based on a single primary emissions factor. 

We should test this and see how it affects our imputed emissions calculations.

grgmiller:
The new process that we are implementing in the 2021 data release will fix this.

CO2 data is considered missing if reported CO2 is zero and fuel consumption is positive.
    1. If a unit has non-missing emissions data for other hours in the same month, calculate a unit-month
        specific EF from the CEMS-reported fuel consumption and emissions
    2. For all remaining missing values, use the subplant and month-specific weighted average emission factor
        from subplant_emission_factors calculated from the EIA-923 data
    3. For any remaining missing values, calculate emissions based on the subplant primary fuel and fuel consumption

PR #259, opened 2022-11-22
Update OGE for to work with PUDL v.2022.11.30 and integrate 2021 data

grgmiller:
This PR closes https://github.com/singularity-energy/open-grid-emissions/issues/258 by updating OGE to work with PUDL v2022.11.30, and integrating 2021 data.

NOTE: We should review and merge https://github.com/singularity-energy/open-grid-emissions/pull/246 into this branch before merging this branch into `development`


grgmiller:
Remaining steps to do:
- [ ] Once Catalyst publishes the data package to Zenodo, update the download link and try running the pipeline from start to finish including the download.
- [ ] Merge #246 
- [ ] Fix issue with negative EFs in consumed outputs
- [ ] Try re-installing environment and make sure it works with the gridemissions solver issue.

grgmiller:
When running the pipeline with the newly-downloaded EIA-930 data, it seems that there may have been some retroactive revisions to the 2020 data, because now when outputting consumed emission factors, we are getting an error message about there being negative emission factors when exporting monthly consumed factors for SEC. The root cause of this issue is described by https://github.com/singularity-energy/open-grid-emissions/issues/214, and was patched by https://github.com/singularity-energy/open-grid-emissions/pull/221 (in which we created a list of BAs with this issue and told the pipeline to use the reported demand values from EIA-930 instead of calculating net demand from generation and interchange). SEC was not previously included in this list, but now seems to be exhibiting these same symptoms.

The larger fix to this is captured by https://github.com/singularity-energy/open-grid-emissions/issues/220, but in the meantime, there are a couple ways we can go about patching this. 
1. Investigate whether there is a new data quality issue in the EIA-930 timeseries for SEC that we need to correct in `eia930.manual_930_adjust()`
2. Add SEC to the `BA_930_INCONSISTENCY` list in `src.consumed()`. However, since this issue seems to be confined to 2020 data for SEC, I'd like to see us update `BA_930_INCONSISTENCY` so that we are only applying this patch for years where it is necessary, rather than applying it to these BAs in all years. One way to do this would be to make `BA_930_INCONSISTENCY` a dictionary, where the key is the year and the value is a list of the BAs that need this patch in that year.

environment.yml
```diff
@@ -15,19 +16,21 @@ dependencies:
   - pip
   - plotly
   - pyarrow
-  - pygeos # used for pudl
+  - pygeos>=0.10,<0.14 # used for pudl
   - pytest
   - python>=3.10,<3.11 # require 3.10.x for compatibility with pudl
   - python-snappy # used for pudl
+  - qdldl-python==0.1.5,!=0.1.5.post2 # used for gridemissions, newer version not working as of 12/12/2022
```
grgmiller, line 23: This is meant to patch https://github.com/singularity-energy/open-grid-emissions/issues/261

grgmiller:
So we are currently running into an issue where the consumed emission calculation in step 18 is returning missing values for all hours in all regions through April of 2021 (for the 2021 year run). I think I've traced the source of this issue to  
`consumed.HourlyConsumed.run()`. Specifically, when we get to the following section of code:
```
# Run
  try:
      consumed_emissions, _ = consumption_emissions(E, G, ID)
  except np.linalg.LinAlgError:
      # These issues happen at boundary hours (beginning and end of year)
      # where we don't have full data for all BAs
      # print(f"WARNING: singular matrix on {date}")
      consumed_emissions = np.full(len(self.regions), np.nan)
```
for the first 2,900 dates in the for loop, the `consumption_emissions()` function is returning the Singular Matrix linalg error that triggers the except clause. Here is the full traceback:
```
File a:\GitHub\open-grid-emissions\notebooks\work_in_progress\../../../open-grid-emissions/src\consumed.py:179, in consumption_emissions(F, P, ID)
    176             # force this to be zero so the linear system makes sense
    177             b[i] = 0.0
--> 179 X = np.linalg.solve(A, b)
    181 for j in perturbed:
    182     if X[j] != 0.0:

File <__array_function__ internals>:180, in solve(*args, **kwargs)

File a:\miniconda3\envs\open_grid_emissions\lib\site-packages\numpy\linalg\linalg.py:400, in solve(a, b)
    398 signature = 'DD->D' if isComplexType(t) else 'dd->d'
    399 extobj = get_linalg_error_extobj(_raise_linalgerror_singular)
--> 400 r = gufunc(a, b, signature=signature, extobj=extobj)
    402 return wrap(r.astype(result_t, copy=False))

File a:\miniconda3\envs\open_grid_emissions\lib\site-packages\numpy\linalg\linalg.py:89, in _raise_linalgerror_singular(err, flag)
     88 def _raise_linalgerror_singular(err, flag):
---> 89     raise LinAlgError("Singular matrix")

LinAlgError: Singular matrix
```

The comment on this code suggests that this error would only be raised when we don't have full data for all of the BAs. However, I'm not sure why there would not be full data now compared to the current release of OGE. It seems like the source of this new issue either has to result from 1) a change to the raw downloaded data from EIA or 2) a change to the way we are processing the data. However, I'm not noticing any changes in the code that would have changed this, so I'm kind of stumped about the cause.

There are a few things related to the cleaning of the 930 data that I'm noticing that I had questions about (may or may not be related to the above issue):
- I notice that the EIA-930 data cleaning for 2021 (step 12 of the data pipeline) is cleaning data all the way back to July 2020. I know that we need some data pre 1/1/2021, but I'm assuming we don't need to go back that far. Would one day (12/31/2020) be sufficient? Further filtering this should in theory at least speed up the data cleaning step by 33%.
- It looks like in step 18, we are loading the cleaned file (eia930_elec.csv), but not necessarily removing the imputed 1 values or filtering to only 2021 values (unless I am missing this). I do note this step:
  ```
  # In some cases, we have zero generation but non-zero transmission
  # usually due to imputed zeros during physics-based cleaning being set to 1.0
  # but sometimes due to ok values being set to 1.0ß
  to_fix = (ID.sum(axis=1) > 0) & (G == 0)
  ID[:, to_fix] = 0
  ID[to_fix, :] = 0
  ```
  but I'm not sure if this is accomplishing the same thing or not

@gailin-p

src/eia930.py
```diff
@@ -173,7 +174,7 @@ def reformat_chalendar(raw):
     print("Filtering")
     cleaned = (
         raw.loc[:, target_cols]
-        .melt(ignore_index=False, value_name="generation")
+        .melt(ignore_index=False, value_name="generation", var_name="variable")
```
grgmiller, line 177: @gailin-p why did you end up adding this argument if it didn't change anything downstream? I'm wondering if there is something here that could be messing with the downstream calculations?

gailin-p: this just guarantees that the column name is what we're splitting in the next line; it doesn't change the function exports. Without this, 'melt' is sensitive to `raw.columns.name`, which can be different depending on where/how the raw data frame was imported

grgmiller:
One other thing I'm noticing about this issue: in 2021, it is affecting all data from 1/1/2021 - 4/30/2021. In 2020, it is only affecting data for the month of March. It's strange that the impact is so neatly cutoff by month, which makes me wonder if there is a clue there - is there some step that we are doing that affects data on a month by month basis?

gailin-p:
The above issue is called by inconsistent transmission interchange and generation in HGMA in 2021 Jan - April. For BAs where there is no generation and no export, the `consumption_emissions` function (from `gridemissions`) perturbs the matrix to make it invertible, however, it doesn't do the same when a BA has zero generation but non-zero export (a situation that's physically impossible and therefore guaranteed not to happen in the `gridemissions` pipeline, where consumed emissions calculations come after physics-based data cleaning). 

OGE generation for HGMA is zero until May 2021: 
Screen Shot 2022-12-23 at 11 05 50 AM

But in both raw and post-gridemissions 930 data, there is non-zero import/export from HGMA in those months: 
Screen Shot 2022-12-23 at 11 09 20 AM
Screen Shot 2022-12-23 at 11 09 35 AM

Crucially, the raw data shows net import to HGMA, which is physically possible with zero generation, but after physics-based cleaning, the 930 data shows net export from HGMA, which is not possible with zero generation. 

There are three easy fixes here, two of which are general (but might also let problem data sneak through in the future), one of which is specific: 
* Expand the perpetuation in the calc to also perturb when a matrix has a zero row 
* Use linalg.lstsq instead of linalg.solve, which is equivalent when there's an exact solution but finds a best fit otherwise 
* Use 930 demand data instead of our generation + 930 interchange for HGMA in the problem months; this is the targeted fix

gailin-p:
As a side-note, because the interchange post physics-based cleaning is >1, it isn't set to zero in our filter. Also, the interchange is 1 in the balance files direct from EIA, so it's not noise introduced by `gridemissions`.

gailin-p:
>I notice that the EIA-930 data cleaning for 2021 (step 12 of the data pipeline) is cleaning data all the way back to July 2020. I know that we need some data pre 1/1/2021, but I'm assuming we don't need to go back that far. Would one day (12/31/2020) be sufficient? Further filtering this should in theory at least speed up the data cleaning step by 33%.

The prior six months is needed for the rolling filter used by `gridemissions` 

>It looks like in step 18, we are loading the cleaned file (eia930_elec.csv), but not necessarily removing the imputed 1 values or filtering to only 2021 values (unless I am missing this).

No need to filter to 2021 values, since we only run the computation on dates in the OGE generation, which is limited to `year`.
We removed imputed ones when we use the 930 data to calculate residual profiles (`eia930.remove_imputed_ones`), but not for the consumed emissions calculation; I think I just overlooked this since the imputed ones were causing major issues with the residual profiles but weren't an issue with the consumed calc. I can add it for consistency.

gailin-p:
> # In some cases, we have zero generation but non-zero transmission
> # usually due to imputed zeros during physics-based cleaning being set to 1.0
> # but sometimes due to ok values being set to 1.0ß
> to_fix = (ID.sum(axis=1) > 0) & (G == 0)
> ID[:, to_fix] = 0
> ID[to_fix, :] = 0

Actually, rewriting this check to make it more general will fix the HGMA bug and seems like the lowest impact option; I'll do that

grgmiller:
This PR should close #163

Issue #263, opened 2022-12-20
Further patching the negative consumed EF issue


grgmiller:
When running the pipeline with the newly-downloaded EIA-930 data, it seems that there may have been some retroactive revisions to the 2020 data, because now when outputting consumed emission factors, we are getting an error message about there being negative emission factors when exporting monthly consumed factors for SEC. The root cause of this issue is described by https://github.com/singularity-energy/open-grid-emissions/issues/214, and was patched by https://github.com/singularity-energy/open-grid-emissions/pull/221 (in which we created a list of BAs with this issue and told the pipeline to use the reported demand values from EIA-930 instead of calculating net demand from generation and interchange). SEC was not previously included in this list, but now seems to be exhibiting these same symptoms.

The larger fix to this is captured by https://github.com/singularity-energy/open-grid-emissions/issues/220, but in the meantime, there are a couple ways we can go about patching this. 
1. Investigate whether there is a new data quality issue in the EIA-930 timeseries for SEC that we need to correct in `eia930.manual_930_adjust()`
2. Add SEC to the `BA_930_INCONSISTENCY` list in `src.consumed()`. However, since this issue seems to be confined to 2020 data for SEC, I'd like to see us update `BA_930_INCONSISTENCY` so that we are only applying this patch for years where it is necessary, rather than applying it to these BAs in all years. One way to do this would be to make `BA_930_INCONSISTENCY` a dictionary, where the key is the year and the value is a list of the BAs that need this patch in that year.

_Originally posted by @grgmiller in https://github.com/singularity-energy/open-grid-emissions/issues/259#issuecomment-1360372679_

PR #265, opened 2022-12-21
Patch negative efs

grgmiller:
Closes https://github.com/singularity-energy/open-grid-emissions/issues/263.

It seems like many of the regions that we had previously flagged as having issues are no longer having issues.


src/consumed.py
```diff
@@ -277,7 +282,7 @@ def output_results(self):
         for ba in self.regions:
             if (ba in self.import_regions) or (ba in self.generation_regions):
```
grgmiller, line 283: @gailin-p what is happening with this line of code? What do we want to have happen if the ba is an import region or generation region? Should the next if statement be an `elif`?

gailin-p: This just skips the rest of the loop for regions that are generation only (self.generation_regions) or regions that are outside the US but trade with US regions (self.import_regions): ie regions where we don't want to export files

PR #266, opened 2022-12-23
Fix consumed calc

gailin-p:
Fix missing values by running check for non-zero export with zero generation: check now happens *after* converting nan to 0. 

Make 930 usage consistent with residual calculation by rounding values < 1.5 to 0; this gets rid of 1 values added by `gridemissions`.

### DIFF
--- a/src/consumed.py
+++ b/src/consumed.py
@@ -210,6 +210,11 @@ class HourlyConsumed:
         # 930 data
         self.eia930 = BaData(eia930_file)
 
+        # Round 930 data to zero to clear out imputed values.
+        # to mimic behavior of eia930.remove_imputed_ones,
+        # round all values with abs(x) < 1.5 to 0
+        self.eia930.df[self.eia930.df.abs() < 1.5] = 0
+
         # Emission factors for non-US bas
         self.default_factors = get_average_emission_factors(prefix, year)
 
@@ -470,6 +475,7 @@ class HourlyConsumed:
     def run(self):
         for pol in POLLUTANTS:
             for adj in ADJUSTMENTS:
+                total_failed = 0
                 col = get_rate_column(pol, adjustment=adj, generated=False)
                 print(f"{pol}, {adj}", end="...")
                 # Calculate emissions
@@ -494,9 +500,13 @@ class HourlyConsumed:
                         except np.linalg.LinAlgError:
                             # These issues happen at boundary hours (beginning and end of year)
                             # where we don't have full data for all BAs
-                            # print(f"WARNING: singular matrix on {date}")
+                            total_failed += 1
                             consumed_emissions = np.full(len(self.regions), np.nan)
 
                     # Export
                     for (i, r) in enumerate(self.regions):
                         self.results[r].loc[date, col] = consumed_emissions[i]
+                if total_failed > 0:
+                    print(
+                        f"Warning: {total_failed} times failed to solve for consumed emissions, {pol} {adj}"
+                    )
--- a/src/consumed.py
+++ b/src/consumed.py
@@ -508,5 +508,5 @@ class HourlyConsumed:
                         self.results[r].loc[date, col] = consumed_emissions[i]
                 if total_failed > 0:
                     print(
-                        f"Warning: {total_failed} times failed to solve for consumed emissions, {pol} {adj}"
+                        f"Warning: {total_failed} hours failed to solve for consumed emissions, {pol} {adj}"
                     )
--- a/src/consumed.py
+++ b/src/consumed.py
@@ -454,14 +454,18 @@ class HourlyConsumed:
             else:
                 G[i] = self.generation.loc[date, r]
 
+        E = np.nan_to_num(E)
+        G = np.nan_to_num(G)
+        ID = np.nan_to_num(ID)
+
         # In some cases, we have zero generation but non-zero transmission
         # usually due to imputed zeros during physics-based cleaning being set to 1.0
-        # but sometimes due to ok values being set to 1.0ß
+        # but sometimes due to ok values being set to 1.0
         to_fix = (ID.sum(axis=1) > 0) & (G == 0)
         ID[:, to_fix] = 0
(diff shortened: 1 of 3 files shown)


src/consumed.py
```diff
@@ -210,6 +210,11 @@ def __init__(
         # 930 data
         self.eia930 = BaData(eia930_file)
 
+        # Round 930 data to zero to clear out imputed values.
+        # to mimic behavior of eia930.remove_imputed_ones,
+        # round all values with abs(x) < 1.5 to 0
+        self.eia930.df[self.eia930.df.abs() < 1.5] = 0
```
grgmiller, line 216: In `eia930.remove_imputed_ones()`, we actually round values to zero decimal places and then replace 1s with zeros, which is almost functionally equivalent to this, except that it wouldn't convert -1 values (which here we are doing by calling `.abs()`. Since in the remove_imputed_ones function we are only dealing with net generation data maybe this shouldn't matter (are there any cases where there might be an imputed -1?) maybe this can be slightly different but just wanted to flag in case we do want this function to be exactly functionally equivalent to the other function.

src/consumed.py
```diff
@@ -454,18 +459,23 @@ def build_matrices(self, pol: str, adj: str, date):
...
+        G = np.nan_to_num(G)
+        ID = np.nan_to_num(ID)
+
         # In some cases, we have zero generation but non-zero transmission
         # usually due to imputed zeros during physics-based cleaning being set to 1.0
-        # but sometimes due to ok values being set to 1.0ß
+        # but sometimes due to ok values being set to 1.0
         to_fix = (ID.sum(axis=1) > 0) & (G == 0)
         ID[:, to_fix] = 0
         ID[to_fix, :] = 0
```
grgmiller, line 471: So in https://github.com/singularity-energy/open-grid-emissions/pull/259#issuecomment-1364105527 you mentioned re-writing this check to fix the HGMA issue but it looks like nothing has changed here - it there another commit you need to push? I'm not sure that I'm clearly seeing in these proposed changes what is fixing the HGMA issue

grgmiller (changes requested):

src/consumed.py
gailin-p: > we actually round values to zero decimal places

I think we should maybe take out the rounding in `remove_imputed_ones` -- seems to me like it throws away info for no reason. For example, even though the raw 930 data doesn't have decimals, if the physics based cleaning were working perfectly, we'd expect it to use info from multiple regions to adjust values and potentially give them meaningful decimal places: for example, if raw export from one region is 10, and import to that region is reported as 11 by its trading partner, we'd want `gridemissions` to set both to 10.5. 

I'm sure it doesn't make a big difference in the results, but it could make a small difference and I don't know what we gain from it.

gailin-p: > wouldn't convert -1 values

I think generation is guaranteed to be positive by `gridemissions`, but we could add a check in `remove_imputed_ones` for consistency

src/consumed.py
gailin-p: Fixed by moving nan to zero conversion above this ID / Gen check

src/consumed.py
gailin-p: propose making `remove_imputed_ones` behavior consistent in [dac1e95](https://github.com/singularity-energy/open-grid-emissions/pull/266/commits/dac1e9551338a555e7aae08ef48f6dd32396573a)

grgmiller (approved):

src/consumed.py
```diff
@@ -314,6 +319,12 @@ def output_results(self):
                     time_dat.datetime_local.dt.year == self.year
                 ]  # keep year of local data
 
+                # if there are small gaps in the consumed data,
+                # linearly interpolate the missing values, up to a maximum of 2 consecutive missing hours
+                time_dat.loc[:, CONSUMED_EMISSION_RATE_COLS] = time_dat.loc[
+                    :, CONSUMED_EMISSION_RATE_COLS
+                ].interpolate(method="linear", axis=0, limit=2, limit_area="inside")
```
grgmiller, line 326: In the hourly consumed calculations, there are certain hours with missing data. I'm not sure if this has always been happening and I've just never noticed, or whether this is a new issue that just started happening. It looks like in the 2020 data posted online, these gaps are less frequent or not occuring at all. I proposed this linear interpolation as a patch, but not sure if we should try and track down the root cause of these missing values? @gailin-p  what do you think?

gailin-p: What's the frequency of the missing hours? If they're much more frequent in the 2021 data then I think we should try to figure out why. 
Also, are the same # of hours missing in the posted 2020 data as in 2020 data computed using the current code? If there are more using the current code, that also seems like a warning sign

grgmiller: Let me look into this but I think the frequency is greater.

grgmiller: Ok the existing consumed emissions data for 2020 posted online has 168 missing values and looks like this when visualized for missing values: 
image
The new consumed emissions data has 5562 missing values, most of which appear to be coming from EEI and SEC (which are coincidentally the only two BAs now listed in `BA_930_INCONSISTENCY`):
image

grgmiller: Testing this for the 2021 data, I'm now discovering that the number of missing values is not stable, and it changes each time the consumed emissions calc is run. Previously, I was seeing a handful of missing values in each file, and in the couple of times I've run this locally this morning, sometimes I am getting no missing values, sometimes I am getting one hour with missing values, sometimes I am getting two hours. This seems like an issue that this isn't stable... does this maybe have to do with how we are perturbing the calculation?

grgmiller: It turns out that this issue was caused by using the MKL build of the `blas` package, as documented in https://github.com/numpy/numpy/issues/20233 and https://github.com/conda/conda/issues/11023. Essentially what was happening was that when we were inverting the matrix, we were getting inconsistent results. This was fixed by changing the environment to include `nomkl` so that the `openblas` version of `blas` is installed.

PR #267, opened 2022-12-28
Release 0.2.0

grgmiller:
To do before merging into main:
 - [x] Merge https://github.com/singularity-energy/open-grid-emissions/pull/259 into `development`
 - [x] Update the `citation.CFF` file with the new version
 - [ ] Update the documentation
 - [x] Test run the pipeline
 - [x] Agree on a release version number (I had originally proposed `0.2.0`, but we might want to save that for a future release and just move to `0.1.3`) @gailin-p any thoughts?
 - [x] Write the release notes
 - [x] Search for all "TODO" and "NOTE" comments and make sure they are addressed.


gailin-p:
On version number, I'd maybe go for 0.2.0 since we're updating to match PUDL's new format, which is a breaking change (new code won't work on old data). But since both PUDL and OGE are in 0.x, breaking changes are to be expected... 

Either 0.2.0 or 0.1.3 is fine with me

src/data_cleaning.py
```diff
@@ -191,12 +172,27 @@ def generate_subplant_ids(start_year, end_year, cems_ids):
     )
 
 
+def manually_update_subplant_id(subplant_crosswalk):
+    """
+    This function corrects subplant mappings not caught by update_subplant_id.
+
+    This is temporary until the pudl subplant crosswalk includes boiler-generator id matches.
```
gailin-p, line 179: Huh! Did this end up being the root issue of the GTN conversion problems with this plant?

test/test_emissions.py
```diff
@@ -7,51 +7,88 @@
...
+            "n2o_mass_lb_adjusted": [0.5, 0.2],
+            "co2_mass_lb_for_electricity": [1.0, 2.0],
+            "ch4_mass_lb_for_electricity": [3.0, 6.0],
+            "n2o_mass_lb_for_electricity": [0.5, 0.2],
+        }
+    )
     return df
 
 
+# NOTE: Why are we testing these function specifically?
```
gailin-p, line 34: Do we need to answer this question?

gailin-p (approved):

src/data_cleaning.py
grgmiller: It did. It's because a missing association between the boiler and generator, so something to fix in pudl in the future.

test/test_emissions.py
grgmiller: Good question. I think @miloknowles had implemented all of our original tests, and maybe this is just where he got started or these were the functions he was using at the time. Perhaps in the future we can work on improving our test coverage, but we also have a lot of validation checks built in.

miloknowles: I think I had added these tests when we originally added those CO2-eq helpers - feel free to delete if you don't think we need them or if they're adding overhead to maintain

grgmiller (commented):
Everything looks good to me

Issue #271, opened 2023-01-20
Missing data due to fuel code mismatch between EIA-860 and EIA-923


grgmiller:
We've noticed large fluctuations in the fleet emission rate for petroleum across each of the years. This issue seems to have existed prior v0.2.0, but we never noticed it before. 

2019: 14,726,036,229.73 lbCO2 / 566,287 MWh = 25,928.32 lb/MWh
2020: 18,206,127,107.72 lbCO2 / 7,427,800 MWh = 2,451.08 lb/MWh
2021: 22,809,325,967 / 28,720,749 = 794 lb/MWh

Steps: 
- Implement check to identify large differences between fleet emission rates across years (how widespread is this issue?)
- This suggests that we should probably implement automated anomaly detection to catch these issues in the future. 

My guess would be that the source of this issue is most likely to be a result of:
- Missing data in CEMS that is not getting flagged for imputation
- An issue with the data crosswalking between EIA and EPA
- Some issue with the gross to net conversion (although no widespread issues were flagged in the pipeline).

grgmiller:
So after digging into this in more detail, this specific issue appears to result from a specific plant, and may be a symptom of several broader issues. 

As we showed above, in 2019, total petroleum fleet emissions were calculated as ~14.7 billion lb. Of this total, approximately 14 billion lb comes from a single plant ("Columbia", id 8023). There are a couple of issues with this:
- Plant 8023 is actually a coal plant (it uses a small amount of DFO for startup)
- While CEMS reports over 5.8 million MWh of generation from this plant in 2019, EIA only reports about 3 thousand MWh of generation.

This suggests that there are two big issues:
1. There is something wrong with our plant primary fuel identification
2. Our gross to net generation conversion is "accurate" in that it is converting 5.8 million MWh to 3,000 MWh, but we should not be allowing this when there is that big of a discrepancy in gross to net. This is a situation where the algorithm should be defaulting to a more reasonable default gross to net ratio. 

What else do we know about what's going on?
- The pipeline is defaulting to use CEMS data for this plant
- The cleaned CEMS data for this plant is being assigned a fuel code of DFO, which is incorrect.

I'm going to poke around with this plant and see what I can find out.

grgmiller:
So it looks like the issue may be with the `pudl.analysis.allocate_net_gen.allocate_gen_fuel_by_generator_energy_source()` module. It appears that it is dropping some data.

[...]

Issue #177, opened 2022-08-16
Update default gross to net generation conversion factors


grgmiller:
According to page 16 of the EIA's Electric Power Monthly [Technical Notes](https://www.eia.gov/electricity/monthly/pdf/technotes.pdf), whenever a generator only reports gross generation, they use the following conversion factors to calculate net generation:
![image](https://user-images.githubusercontent.com/45949268/184950232-a2bc503a-e9b8-4013-8491-94c0843ce0db.png)

Instead of defaulting to setting net generation equal to gross generation when we have no other information, we should use these as default factors. 

However, it is unclear whether the prime mover factors and environmental equipment factors are used in combination (i.e. a combined cycle steam plant with flue gas desulfurization would use (0.97 x 0.97 = 0.94)

grgmiller:
As part of https://github.com/singularity-energy/open-grid-emissions/pull/273, we started using these default prime-mover based factors as the backstop gross to net generation ratio if no other methods worked. However, there are several ways this still needs to be improved:
- [ ] Add the impact of environmental equipment to the GTN ratio
- [ ] Double check that all prime movers are covered, and identify default gtn factors for those that are not.

PR #273, opened 2023-01-25
Fix bug with EIA-923 Generation and Fuel Allocation

grgmiller:
This PR closes https://github.com/singularity-energy/open-grid-emissions/issues/271, fixing an issue where our data pipeline was dropping a substantial amount of data due to mismatches in the reported energy source codes used in EIA-860 and EIA-923, and our failure to validate the outputs of the allocation process more carefully. While fixing this issue, we also came across several other issues that were causing anomalous emission factor outputs.

## Fixes the generation and fuel allocation process
- Fix bugs in pudl allocate_net_gen module, as described in this PR (https://github.com/catalyst-cooperative/pudl/pull/2235):
> - Adds a new function `add_missing_energy_source_codes_to_gens()` that adds energy_source_codes that appear in the `gf` table but not `gens` to `gens`. 
> - In some cases, non-zero fuel consumption and net generation is reported in the EIA-923 generation and fuel table that is associated with an energy_source_code that is not associated with that plant-prime mover in the gens table, which would cause these data to get dropped when these two tables are merged. To fix this, for each plant-pm, this function identifies such esc, and adds them to the `gens_at_freq` table as new energy_source_code columns.
>- The sub-function `identify_missing_gf_escs_in_gens()`  identifies when there are fuels reported in the gf table for that plant-pm that are not listed in the gens table for that plant pm. 
> - Adds the `MISSING_SENTINEL` value to the `net_generation_mwh_g_tbl` column. For some reason, this column had been commented out, which was leading to NaNs appearing in the data when dividing by this column when the value was zero. I un-commented this line.
> - In `allocate_net_gen_by_gen_esc()`, we no longer allow `frac_from_g_tbl` to be greater than 100%. This was previously happening when the mwh reported in the g table were greater than the mwh reported in the gf table. However, numbers greater than 100% was causing the `frac_missing_from_g_tbl` to become negative, which was resulting in nonsensical allocations. We implement the same cap on `frac_from_bf_tbl` in `allocate_fuel_by_gen_esc()`.
> - In  `allocate_fuel_by_gen_esc()`, when calculating `frac_cap`, the code had been dividing `capacity_mw` by `capacity_mw_unit_fuel`. However, this was resulting in some nonsensical allocations because fuel is being allocated by PM-fuel, not by unit. We changed this to divide by `capacity_mw_pm_fuel` instead. This is consistent with how `frac_cap` is calculated in the `allocate_net_gen_by_gen_esc()` function
> - Rename `adjust_energy_source_codes()` to `adjust_msw_energy_source_codes()` to more precisely describe what the function does
- Adds new entries to the manual emissions factor tables for NOx and SO2 to represent fuel-boiler combinations that had previously been getting dropped from the data due to this bug.
- Changes the pudl version that we use in our environment from `catalyst-cooperative/pudl@main` to `grgmiller/pudl@oge_release`. This will give us more control over performing fixes like this in the future. 
- Adds a new validation check to the EIA-923 data cleaning process to verify that for each plant, the total allocated fuel and generation matches the total fuel and generation reported in the input generation and fuel table (basically that the allocation process is not dropping or inflating the data).

## Updates plant primary fuel identification
- When assigning the plant primary fuel based on the most consumed fuel, we were previously assigning this based on the fuel with the highest `fuel_consumed_mmbtu`. However, we should be using `fuel_consumed_for_electricity_mmbtu` since we want to assign the primary fuel used for electricity generation. 
- Sometimes nuclear generators report 0 fuel consumption in EIA-923. Since we were assigning a plant's primary fuel first based on fuel consumption, this meant that sometimes if a nuclear plant had a backup fossil generator, the plant was being assigned the fuel code of that backup generator. To fix this, we now assign the primary fuel of any plant that contains a nuclear unit based on the nameplate capacity of the unit.

## Updates to gross to net generation conversions
- Previously, when converting CEMS gross generation to net generation, we had filtered out any ratios that were greater than 1.5 or less than 0.2. However, these values were somewhat arbitrary and turning out to be too wide of a range. For example, when there was a large discrepancy between CEMS gross generation and EIA-923 net generation, we were scaling the CEMS generation to match, even though the fuel consumption and emissions reported in CEMS also disagreed and were not being scaled. This was leading to instances where a plant was using CEMS CO2 totals but EIA-923 net generation totals, resulting in the plant having abnormally high emission rates. To be consistent, if we are going to use CEMS data at all, we want to make sure that the net generation values are reasonable given the reported net generation. After analyzing three years (2019-2021) of annual gross to net ratios, both at the plant and subplant levels, it appears that generally the interquartile range of GTN ratios is between 0.75 and 1.00, with an upper bound around 1.25. Thus, we are now using 0.75 as the lower bound for filtering out ratios, and 1.25 as the upper bound. 
- Previously, the backstop gross to net generation approach if all other conversion factors were not available was to assume that gross generation equaled net generation (i.e. a GTN ratio of 1). However, as identified in https://github.com/singularity-energy/open-grid-emissions/issues/177, the EIA has default gross to net conversion factors for each prime mover that they use. This PR introduces these PM-specific gross to net conversion factors as the default backstop option now. As noted in the issue, there are still improvements that need to be made before #177 can be closed, but this is a step in the right direction.

## Flags potentially anomalous generated emission factors
- When outputting annual plant level data, we add a new validation check that calculates a generated co2 rate, and flags any rates that appear to be anomalous, so that we can manually inspect these plants to see if there are any unexpected results. For this test, we define anomalous values two ways. On the high end, the check flags the plant if the co2 rate is higher than 15,000 lb/MWh. On the low end, the check flags any plants that have rate lower than 10 lb/MWh but higher than 0 MWh.

## Other
- Fixes futurewarnings from pandas related to grouping non-numeric columns when using pd.groupby(). This involved only summing numeric columns and sometimes adding the `numeric_only=True` argument to .sum(). 

TODO:
- [x] remove WIP notebooks
- [ ] Update documentation


notebooks/work_in_progress/test_pudl_allocate_net_gen.ipynb
```diff
@@ -0,0 +1,724 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
```
gailin-p, line 8: Do we have any policy on when we want to check in notebooks in the WIP folder to the main branch? I think ideally we'd move notebooks we want to keep out of WIP and into manual data or another folder when a PR is ready to merge into main

src/data_cleaning.py
```diff
@@ -716,6 +721,15 @@ def calculate_aggregated_primary_fuel(
...
     ].fillna(agg_primary_fuel["primary_fuel_from_mode"])
 
+    # sometimes nuclear generators report 0 fuel consumption in EIA-923.
+    # To ensure that a nuclear plant does not get assigned a fuel code of
+    # a backup generator, we use the nameplate capacity to assign the
+    # primary fuel for any plants that contain a nuclear generator
+    agg_primary_fuel.loc[
+        agg_primary_fuel["primary_fuel_from_capacity_mw"] == "NUC",
+        f"{level}_primary_fuel",
+    ] = "NUC"
```
gailin-p, line 731: Is this also the case for any other generators that don't use a fossil fuel as their primary? (ie, hydro, wind, solar)

src/gross_to_net_generation.py
```diff
@@ -45,12 +45,15 @@ def convert_gross_to_net_generation(cems, eia923_allocated, plant_attributes, ye
...
                 "annual_fuel_ratio",
+                "default_gtn_ratio",
             ]
         ],
         how="left",
         on=["plant_id_eia", "subplant_id", "report_date"],
         validate="m:1",
     )
+    # if there is a missing default gtn ratio, fill with 0.97
+    cems["default_gtn_ratio"] = cems["default_gtn_ratio"].fillna(0.97)
```
gailin-p, line 56: Should we raise a warning here? Shouldn't all defaults come from  data/manual/default_gross_to_net_ratios?

src/gross_to_net_generation.py
```diff
@@ -367,16 +379,33 @@ def calculate_gross_to_net_conversion_factors(
...
 
+    # add default GTN ratios from EIA
+    gtn_conversions = gtn_conversions.merge(
+        load_data.load_default_gtn_ratios(),
+        how="left",
+        on="prime_mover_code",
+        validate="m:1",
+    )
+    # for any misisng default gtn ratios, use a ratio of 0.97
+    gtn_conversions["default_gtn_ratio"] = gtn_conversions["default_gtn_ratio"].fillna(
```
gailin-p, line 390: This default-filling logic occurs twice (here and in convert_gross_to_net_generation). Should only be in one place. Also, as commented above, maybe we should throw a warning instead of giving a default value?

src/gross_to_net_generation.py
```diff
@@ -436,15 +485,19 @@ def filter_gtn_conversion_factors(gtn_conversions):
         "annual_plant_ratio",
         "annual_fuel_ratio",
     ]:
+        # remove any factors that would scale net generation to less than 75% of gross generation
+        # In general, the IQR of GTN ratios is between 0.75 and 1, with an upper bound around 1.25
         # remove any ratios that are negative to avoid flipping the shape of the profile
-        factors_to_use.loc[factors_to_use[scaling_factor] < 0, scaling_factor] = np.NaN
+        factors_to_use.loc[
+            factors_to_use[scaling_factor] < 0.75, scaling_factor
+        ] = np.NaN
```
gailin-p, line 493: Do these incorrect values remain in an output column somewhere so we can see where we're overwriting data-derived values with defaults? Might be useful for IDing other issues.

data/manual/default_gross_to_net_ratios.csv
```diff
@@ -0,0 +1,14 @@
+prime_mover_code,default_gtn_ratio
```
gailin-p, line 1: Can we add the default/fallback value (.97) to this file so it's easier to find?

src/validation.py
```diff
@@ -40,6 +40,35 @@ def validate_year(year):
...
+        ]
+    ].sum()
+    # calculate the difference between the values
+    plant_total_diff = plant_total_gf - plant_total_alloc
+    # flag values where the absolute difference is greater than 10 mwh or mmbtu
+    mismatched_allocation = plant_total_diff[
+        (abs(plant_total_diff["fuel_consumed_mmbtu"]) > 10)
+        | (abs(plant_total_diff["net_generation_mwh"]) > 10)
+    ]
+    if len(mismatched_allocation) > 0:
```
gailin-p, line 67: Are there currently any plants where this is an issue?

gailin-p (approved):

notebooks/work_in_progress/test_pudl_allocate_net_gen.ipynb
grgmiller: I was planning on deleting this workbook before merging the PR. But in general, I think that we should have a policy that any notebook in WIP has to be associated with an active issue/PR, otherwise it needs to be deleted or moved to another directory.

Maybe though I should consider keeping this notebook since there are still some plants where the allocation isn't matching up completely.

src/data_cleaning.py
grgmiller: Good question. I did look at this for wind and hydro and didn't see any issues - they all report fuel consumption data. 

There are a number of entries in the EIA-923 generation and fuel table that report positive net generation but zero fuel consumption. Many of these are for `CA` prime movers, which means the steam portion of a combined cycle plant, which makes sense because the fuel is consumed at the `CT` prime mover. Ignoring those, the majority of the remaining data is nuclear steam turbines. There are a few other edge cases that wouldn't be caught by this step, but it's small enough that I don't think it is worth our time to address it now. I will however create an issue to investigate this in the future.

src/gross_to_net_generation.py
grgmiller: Yes good point. It would be good to flag where there are missing prime movers in the default gross to net ratios table

src/gross_to_net_generation.py
grgmiller: Good catch. I think that we want to throw a warning but also probably use the default fill number for now, just so that we don't get any NA values when doing the multiplication. But I agree we should only be doing this filling in a single spot.

data/manual/default_gross_to_net_ratios.csv
grgmiller: I'll have to add entries for each prime mover, and maybe just add a column that indicates the data source.

src/validation.py
grgmiller: There are, but none of the magnitudes looked huge, and I don't think we have the bandwidth to track down the issues with each individual plant now. 

For example, here is the output for 2021:

```
WARNING: Allocated EIA-923 doesn't match input data for plants:
              net_generation_mwh  fuel_consumed_mmbtu  fuel_consumed_for_electricity_mmbtu
plant_id_eia                                                                              
1206                   60.568183          1241.475753                          1241.475753
1316                  -48.000000             0.000000                             0.000000
52148                   1.601000            18.000000                             7.000000
54224                  14.960000           150.000000                            66.000000
54967                   1.960000            24.000000                            24.000000
58256                -241.000000             0.000000                             0.000000
59260                 836.856000         13729.000000                         13729.000000
59825                9539.000000         84354.000000                         84354.000000
```
Looking at the two plants at the bottom, the missing data from 59260 is because the plant retires mid year (according to EIA-860), but there is still reported fuel consumption after that. Plant 59825 is a solar plant that has some wind generation reported in the gf table. These are things that we should look into, but they probably requiring tracking a lot of specific edge cases, and none of the magnitudes are that great. 


However, maybe we should consider exporting a "warnings.txt" file that saves all of the outputs of the warning messages for each run so that we can easily track the data quality issues that were flagged but ignored for each run.

src/gross_to_net_generation.py
grgmiller: Yes, these remain in the `gross_to_net_conversions_{year}.csv` file

PR #274, opened 2023-01-31
add eGRID2021 data

grgmiller:
- Adds eGRID 2021 data to our list of downloads
- Updates the list of non-grid connected plants based on additions to the list in eGRID 2021.
- Fixes an issue where a plant with no reported ba_code was getting filled with the incorrect code based on the ba_name. 

In comparing our 2021 data to the eGRID 2021 data, I found that there are some plants that EIA-860 identifies as being in ISNE that are getting assigned to NYIS by pudl, and thus are categorized in a different BA than they are in eGRID. All of these plants seem to be physically located in the state of New York, but are listed with an ISNE BA code. Also, all of these plants are pretty small. See: https://github.com/catalyst-cooperative/pudl/issues/2255. For now, I was going to leave this in place until this is resolved on the pudl side.

### DIFF
--- a/src/data_cleaning.py
+++ b/src/data_cleaning.py
@@ -1996,6 +1996,15 @@ def create_plant_ba_table(year):
         "balancing_authority_code_eia"
     ].fillna(value=np.NaN)
 
+    # load the ba name reference
+    ba_name_to_ba_code = pd.read_csv(manual_folder("ba_reference.csv"))
+    ba_name_to_ba_code = dict(
+        zip(
+            ba_name_to_ba_code["ba_name"],
+            ba_name_to_ba_code["ba_code"],
+        )
+    )
+
     # specify a ba code for certain utilities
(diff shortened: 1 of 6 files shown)


src/data_cleaning.py
```diff
@@ -2006,6 +2015,9 @@ def create_plant_ba_table(year):
     )
 
     # fill missing BA codes first based on the BA name, then utility name, then on the transmisison owner name
+    plant_ba["balancing_authority_code_eia"] = plant_ba[
```
gailin-p, line 2018: how big is the impact of this change on previously-released data? I assume it's pretty small, just would be nice to know for the release notes

gailin-p (approved):

src/data_cleaning.py
grgmiller: There is a single plant that was previously assigned to BPAT that will now be assigned correctly to GRID in 2021

Issue #279, opened 2023-02-11
CEMS data cleaning is dropping some CEMS data


grgmiller:
When looking into BA-level discrepancies between eGRID2021 data and OGE data, I noticed that the total CO2 emissions for the SOCO balancing authority were about 20% lower than the total reported in eGRID. Comparing plant-level results within SOCO, I noticed that several plants, including "Barry" (ID 3) had substantially less emissions reported in OGE than eGRID. Comparing unit-level CO2 totals for Barry in eGRID2021 and our cleaned CEMS file revealed that our CO2 totals for some units were lower than those in eGRID. These data were both coming from CEMS, so it was puzzling why the totals would be different. I tried loading the "raw" CEMS data from pudl (ie before we apply any data cleaning steps in the OGE pipeline), and these totals matched those in eGRID, which suggests that there is an issue somewhere in `data_cleaning.clean_cems()` that may be dropping or altering data in an unexpected way. 

After stepping through the data cleaning process, it appears that there may be several separate issues:
1. During `emissions.fill_cems_missing_co2()` there are several unexpected things happening: 1) co2 emissions are being added to unit 1, which has 0 fuel consumption or generation. 2) The "filled" emissions for unit 7B are now lower than the original emissions. This function performs several operations based on the index of the dataframe, and I am wondering if the index values are shifting at some point. 
2. Then, `data_cleaning.remove_cems_with_zero_monthly_data()` is removing co2 data (but not generation or fuel data), likely from those records for which co2 was added for hours of zero fuel consumption in the previous step.

grgmiller:
It turns out that when you merge two dataframes together, the original index of the left dataframe is not preserved unless you are merging on the index. In `emissions.fill_cems_missing_co2()`, when we were merging emisison factors into the `missing_co2` dataframe, the index was being reset. Then when we updated the cems dataframe using the updated values, since `pd.update` matches based on the index, the co2 values were overwriting data for the incorrect plant/hour. This was likely affecting hundreds of thousands of rows of data (out of several million) so this bug was having a pretty substantial impact - basically overwriting hourly co2 data with co2 data from other plants and hours. 

How can we prevent this from happening again? Whenever we are using `pd.update` or merging on indexes, we need to make sure that the index of the right dataframe has not been accidentally reset. One way to do this is flag if the right dataframe contains an index value of `0` and an index value of `len(df)-1`, which indicates that the index has been reset. 
[...]

PR #280, opened 2023-02-11
Fix bug with CEMS missing CO2 data filling

grgmiller:
This PR closes https://github.com/singularity-energy/open-grid-emissions/issues/279. 

Updates:
- Fixes issue where the index of a dataframe that was being used to update the cems dataframe was being reset, causing the wrong data to be overwritten, and leading to incorrect CEMS output data.
- Adds validation check to ensure that non-missing CEMS CO2 data is not being modified by `emissions.fill_cems_missing_co2()`
- Adds a validation check to ensure that none of the data being removed from CEMS by `data_cleaning.remove_cems_with_zero_monthly_data()` actually has non-zero data associated with it.
- Updates the eGRID validation notebook to more helpfully format the BA-total comparison table

### DIFF
--- a/src/validation.py
+++ b/src/validation.py
@@ -265,6 +265,38 @@ def test_for_missing_energy_source_code(df):
     return missing_esc_test
 
 
+def check_non_missing_cems_co2_values_unchanged(cems, year):
+    """Checks that no non-missing CO2 values were modified during the process of filling."""
+    print("    Checking that original CO2 data in CEMS was not modified by filling missing values...", end="")
+    # re-load the raw cems data
+    cems_original = load_data.load_cems_data(year)
+    # only keep non-zero and non-missing co2 values, since these should have not been modified
+    cems_original = cems_original.loc[
+        cems_original["co2_mass_lb"] > 0,
+        ["plant_id_eia", "emissions_unit_id_epa", "datetime_utc", "co2_mass_lb"],
+    ]
+    test_fill = cems_original.merge(
+        cems[["plant_id_eia", "emissions_unit_id_epa", "datetime_utc", "co2_mass_lb"]],
+        how="left",
+        on=["plant_id_eia", "emissions_unit_id_epa", "datetime_utc"],
+        validate="1:1",
+        suffixes=("_original", "_postfill"),
+    )
+    test_fill["diff"] = (
+        test_fill["co2_mass_lb_postfill"] - test_fill["co2_mass_lb_original"]
+    )
+    if len(test_fill[test_fill["diff"] != 0]) > 0:
+        print(" ")
+        print(
+            f"WARNING: There are {len(test_fill[test_fill["diff"] != 0])} non-missing CO2 CEMS records that were modified by `fill_cems_missing_co2` in error"
+        )
+    else:
+        print("OK")
+
+    del cems_original
+
+
+
 def test_for_missing_subplant_id(df):
     """Checks if any records are missing a `subplant_id`."""
     print("    Checking that all data has an associated `subplant_id`...  ", end="")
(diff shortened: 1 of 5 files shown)


grgmiller:
After fixing this and comparing the annual plant level data to the v0.2.1 data, it looks like this bug substantially affected 32 plants spread across 10 BAs (AEC, CPLE, ERCO, ISNE, MISO, NYIS, PJM, SOCO, SWPP, and TVA). Most of these plants (14) were in SOCO, but there were also 5 plants in MISO and 5 plants in SWPP that were affected. This error was resulting in about 65 billion lb of CO2 missing from the final results, or about 2% of the total CO2 for electricity.

grgmiller:
So with the previous data before fixing this, this is what the comparison to eGRID BA-level data looked like for the regions with the largest disrepancies:

```
net_generation_mwh	fuel_consumed_mmbtu	fuel_consumed_for_electricity_mmbtu	co2_mass_lb	generated_co2_rate_lb_per_mwh	num_plants
SOCO	0.00	0.01	0.01	-0.19	-0.19	6.0
HST	0.01	0.00	0.00	-0.13	-0.14	0.0
PGE	0.23	0.09	0.09	0.12	-0.08	0.0
IID	0.00	0.00	0.00	-0.08	-0.08	0.0
HECO	0.19	0.18	0.17	0.14	-0.04	11.0
CISO	0.04	-0.00	-0.01	0.01	-0.03	7.0
PACW	0.00	-0.01	-0.01	-0.02	-0.02	0.0
...
TEPC	-0.23	-0.27	-0.27	-0.21	0.03	0.0
TVA	0.00	-0.00	-0.00	0.03	0.03	0.0
TEC	0.01	0.06	0.06	0.06	0.04	0.0
PSEI	0.00	0.04	0.02	0.05	0.05	0.0
AZPS	0.00	0.07	0.07	0.05	0.05	0.0
SRP	0.07	0.17	0.17	0.15	0.08	3.0
TAL	0.00	0.09	0.09	0.09	0.09	0.0
AEC	0.00	0.00	0.00	0.27	0.27	0.0
```

After fixing this, the comparison looks like:

```
       net_generation_mwh  fuel_consumed_mmbtu  fuel_consumed_for_electricity_mmbtu  co2_mass_lb  generated_co2_rate_lb_per_mwh  num_plants
HST                  0.01                 0.00                                 0.00        -0.13                          -0.14         0.0
PGE                  0.23                 0.09                                 0.09         0.12                          -0.08         0.0
IID                  0.00                 0.00                                 0.00        -0.08                          -0.08         0.0
HECO                 0.19                 0.18                                 0.17         0.14                          -0.04        11.0
CISO                 0.04                -0.00                                -0.01         0.01                          -0.03         7.0
PACW                 0.00                -0.01                                -0.01        -0.02                          -0.02         0.0
...
TEPC                -0.23                -0.27                                -0.27        -0.21                           0.03         0.0
TEC                  0.01                 0.06                                 0.06         0.06                           0.04         0.0
PSEI                 0.00                 0.04                                 0.02         0.05                           0.05         0.0
AZPS                 0.00                 0.07                                 0.07         0.05                           0.05         0.0
SRP                  0.07                 0.17                                 0.17         0.15                           0.08         3.0
TAL                  0.00                 0.09                                 0.09         0.09                           0.09         0.0
```

So the worst issues on either end of the scale (SOCO and AEC) have been fixed, but this specific bug was not affecting some of the other, smaller discrepancies in regions like HST, PGE, AZPS, SRP, TAL. These might be caused by other issues.

src/validation.py
```diff
@@ -265,6 +265,55 @@ def test_for_missing_energy_source_code(df):
     return missing_esc_test
 
 
+def check_non_missing_cems_co2_values_unchanged(cems_original, cems):
```
gailin-p, line 268: As we write more validation checks, I think we should separate them into different source files based on either type of validation or where in the pipeline the validation happens. I'd like to avoid getting another 1000+ line file (eg, data_cleaning.py) for two reasons: 1) makes it difficult for to know where to look for code -- you have to either track down a function from data_pipeline.py or know what it's called in order to find it; and 2) when we eventually package OGE, having sensible divisions of logic between files will by default result in useful module names and sensible divisions of logic between modules

gailin-p (approved):

src/validation.py
grgmiller: Good point. I'll add a ticket for this. Maybe we can split this up once we've written more tests so that we can identify what makes sense as a division?

PR #285, opened 2023-02-21
Make PUDL logging show up, and set up logging in OGE

miloknowles:
This PR configures logging in `logging_util.py` such that:
- PUDL logging messages are not hidden
- OGE logging messages are formatted similar to PUDL (see below)
- All logging messages from the data pipeline (including PUDL and other imports) are directed to a logfile (`data/outputs/data_pipeline.log`). The logfile could easily be changed to a different filepath.

Here's an example of running the data pipeline with the new logging system:
Screen Shot 2023-02-21 at 5 02 21 PM

**Note:**
Unfortunately, if you just do `logger.info(df)`, the header won't line up well with the data. So I replaced all of the dataframe printouts to be `logger.info("\n" + df.to_string())`, which will make everything look nice.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,13 +3,14 @@ data/*
 
 example/.ipynb_checkpoints/
 test/__pycache__/
+test/*.txt
 src/__pycache__/
 
 CHANGELOG.md
 
 notebooks/visualization/outputs/*
 
-# Python 
+# Python
 notebooks/.ipynb_checkpoints
 notebooks/*/.ipynb_checkpoints
 .hypothesis/
--- a/src/column_checks.py
+++ b/src/column_checks.py
@@ -17,6 +17,9 @@ those files or uses, then remove it here.
 After any change, re-run data_pipeline to regenerate all files and re-run these
 checks.
 """
+from logging_util import get_logger
+logger = get_logger(__name__)
+
 
 COLUMNS = {
     "eia923_allocated": {
@@ -348,8 +351,8 @@ def check_columns(df, file_name):
     # Check for extra columns. Warning not exception
     extras = cols - expected_cols
     if len(extras) > 0:
-        print(
-            f"WARNING: columns {extras} in {file_name} are not guaranteed by column_checks.py"
+        logger.warning(
+            f"columns {extras} in {file_name} are not guaranteed by column_checks.py"
         )
 
     # Raise exception for missing columns
@@ -464,8 +467,8 @@ def apply_dtypes(df):
         if (col not in dtypes) and (col not in datetime_columns)
     ]
     if len(cols_missing_dtypes) > 0:
-        print(
-            "WARNING: The following columns do not have dtypes assigned in `column_checks.get_dtypes()`"
+        logger.warning(
+            "The following columns do not have dtypes assigned in `column_checks.get_dtypes()`"
         )
-        print(cols_missing_dtypes)
+        logger.warning(cols_missing_dtypes)
     return df.astype({col: dtypes[col] for col in df.columns if col in dtypes})
--- a/src/data_cleaning.py
+++ b/src/data_cleaning.py
@@ -545,14 +545,12 @@ def update_energy_source_codes(df):
         (df["energy_source_code"] == "OTH") & (df["fuel_consumed_mmbtu"] > 0)
     ]
     if len(plants_with_other_fuel) > 0:
-        logger.warning(
-            "After cleaning energy source codes, some fuel consumption is still associated with an 'OTH' fuel type."
+        logger.warning(f"""
+            After cleaning energy source codes, some fuel consumption is still associated with an 'OTH' fuel type.
+            This will lead to incorrect emissions calculations.
+            Check the following plants: {list(plants_with_other_fuel.plant_id_eia.unique())}
+            Assign a fuel type in `data_cleaning.update_energy_source_codes`"""
         )
-        logger.warning("This will lead to incorrect emissions calculations.")
-        logger.warning(
-            f"Check the following plants: {list(plants_with_other_fuel.plant_id_eia.unique())}"
-        )
-        logger.warning("Assign a fuel type in `data_cleaning.update_energy_source_codes`")
 
     return df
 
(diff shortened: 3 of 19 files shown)


src/emissions.py
```diff
@@ -488,7 +490,7 @@ def calculate_nox_from_fuel_consumption(
                     "prime_mover_code",
                     "generator_id",
                 ]
-            ].drop_duplicates()
+            ].drop_duplicates().to_string()
```
grgmiller, line 493: How will this render a dataframe? Will it still be readable as a string?

src/emissions.py
```diff
@@ -687,13 +689,13 @@ def calculate_generator_nox_ef_per_unit_from_boiler_type(
         )
     )
     if len(missing_nox_efs) > 0:
-        print(" ")
-        print(
-            "WARNING: After filling with PM-fuel factors, NOx emission factors are still missing for the following boiler types. An emission factor of zero will be used for these boilers."
+        logger.warning(" ")
```
grgmiller, line 692: Not sure that we need to keep these blank lines or not

src/validation.py
```diff
@@ -207,84 +210,76 @@ def test_for_missing_values(df, small: bool = False):
...
         if small:
-            print(
+            logger.warning(
                 " Found missing values during small run, these may be fixed with full data"
             )
         else:
-            print("!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!")
-            print("WARNING: The above missing values are errors and must be fixed")
-            print("!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!")
+            logger.warning("!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!")
```
grgmiller, line 224: This might be dramatic - we could probably get rid of the exclamation points

src/validation.py
```diff
@@ -326,22 +321,19 @@ def validate_gross_to_net_conversion(cems, eia923_allocated):
...
-            f"WARNING: There are {len(cems_net_not_equal_to_eia)} plants where calculated annual net generation does not match EIA annual net generation."
+        logger.warning(" ")
+        logger.warning(
+            f"There are {len(cems_net_not_equal_to_eia)} plants where calculated annual net generation does not match EIA annual net generation."
         )
-        print(cems_net_not_equal_to_eia)
+        logger.warning("\n" + cems_net_not_equal_to_eia.to_string())
     else:
-        print("OK")
+        logger.info("OK")
```
grgmiller, line 330: Not sure if printing these "OK" messages is necessary or not, or even if we need to print that we are running the test. Should we just print a warning if the warning is triggered? What do you think?

grgmiller (commented):
Thanks @miloknowles this is really helpful. I have a few comments that I wanted to you address or respond to before approval.

.gitignore
```diff
@@ -3,13 +3,14 @@ data/*
 
 example/.ipynb_checkpoints/
 test/__pycache__/
+test/*.txt
```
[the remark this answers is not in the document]
miloknowles: Nope, logs are exported to `data/outputs/data_pipeline.log` (which is already ignored since we ignore `outputs`). A logging test script that I wrote writes logs to `test` though.

src/column_checks.py
```diff
@@ -348,8 +351,8 @@ def check_columns(df, file_name):
     # Check for extra columns. Warning not exception
     extras = cols - expected_cols
     if len(extras) > 0:
-        print(
-            f"WARNING: columns {extras} in {file_name} are not guaranteed by column_checks.py"
+        logger.warning(
```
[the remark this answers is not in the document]
miloknowles: Yep, that's correct (unless I missed some)

src/data_cleaning.py
```diff
@@ -543,14 +545,14 @@ def update_energy_source_codes(df):
...
     ]
     if len(plants_with_other_fuel) > 0:
-        print(
-            "WARNING: After cleaning energy source codes, some fuel consumption is still associated with an 'OTH' fuel type."
+        logger.warning(
+            "After cleaning energy source codes, some fuel consumption is still associated with an 'OTH' fuel type."
         )
-        print("This will lead to incorrect emissions calculations.")
-        print(
+        logger.warning("This will lead to incorrect emissions calculations.")
```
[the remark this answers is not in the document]
miloknowles: Yep, that's possible! Just changed

src/emissions.py
miloknowles: The dataframe will show up in a nice format with rows and columns aligned

src/validation.py
miloknowles: I think have an explicit "this test was OK" is kind of nice, so I kept that

.gitignore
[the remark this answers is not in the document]
grgmiller: Ok got it - this is for tests

grgmiller (approved):

src/validation.py
```diff
@@ -131,29 +134,27 @@ def test_for_negative_values(df, small: bool = False):
     for column in columns_to_test:
         negative_test = df[df[column] < 0]
         if not negative_test.empty:
-            print(" ")
-            print(
-                f"WARNING: There are {len(negative_test)} records where {column} is negative."
+            logger.warning(" ")
```
grgmiller, line 137: @miloknowles could you quickly go through and remove any of these blank logger lines? This was a formatting thing to make the `print("", end="")` thing work, but is no longer necessary since we are just having the "OK" message show up on a new line instead of the same line as the "Checking...." message

grgmiller (changes requested):

src/validation.py
miloknowles: Done

grgmiller (approved):

grgmiller (approved):

grgmiller:
See https://github.com/singularity-energy/open-grid-emissions/pull/286, https://github.com/singularity-energy/open-grid-emissions/pull/289, and https://github.com/singularity-energy/open-grid-emissions/pull/291 for additional tweaks to this PR

PR #286, opened 2023-02-24
Finish logger conversion

grgmiller:
This finishes up the work started in https://github.com/singularity-energy/open-grid-emissions/pull/285, converting a few more print statements that we did not originally catch to logger.info or logger.warning.

### DIFF
--- a/src/download_data.py
+++ b/src/download_data.py
@@ -41,7 +41,9 @@ def download_helper(
     # If the file already exists, do not re-download it.
     final_destination = output_path if output_path is not None else download_path
     if os.path.exists(final_destination):
-        logger.info(f"    {final_destination.split('/')[-1]} already downloaded, skipping.")
+        logger.info(
+            f"    {final_destination.split('/')[-1]} already downloaded, skipping."
+        )
         return False
 
     # Otherwise, download to the file in chunks.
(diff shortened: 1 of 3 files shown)


src/download_data.py
```diff
@@ -109,18 +109,11 @@ def download_pudl_data(zenodo_url: str):
...
-    total_size_in_bytes = int(r.headers.get("content-length", 0))
     block_size = 1024 * 1024 * 10  # 10 MB
-    downloaded = 0
+    logger.info("    Downloading PUDL data...")
     with open(downloads_folder("pudl.tgz"), "wb") as fd:
         for chunk in r.iter_content(chunk_size=block_size):
-            print(
-                f"    Downloading PUDL. Progress: {(round(downloaded/total_size_in_bytes*100,2))}%   \r",
-                end="",
-            )
```
miloknowles, line 120: We may want to keep the PUDL download progress update here, just because the download takes so long. It's nice to know that something is still happening and that your computer isn't frozen or something.

miloknowles (approved):

src/download_data.py
grgmiller: We'd have to keep this as a print statement though, rather than a logging statement, which would be the only print statement in the entire pipeline. Perhaps we can just add a message to the logging statement saying that the file is large and download takes a long time?

miloknowles: I don't think having a `print` statement is a problem. It'll show up for developers watching their terminal, but not in the logs (which is totally fine since it's just a progress update).

grgmiller: Ok added this back in

PR #288, opened 2023-02-24
Small improvements to command line arguments

miloknowles:
I was originally debugging the `--small` option for the data pipeline, which crashes at the EIA-930 cleaning step. During that process, I made a couple improvements to how command line arguments are handled in `data_pipeline.py`:
- Boolean flags like `--small`, `--flat`, and `--skip_outputs` don't need a value after them any more. If you omit these arguments, the default is used. If you include `--small`, that value will be true. This is to be more consistent with other command line interfaces.
- I added a step to print out the command line arguments that are being used, so that they're clearly shown to the user and captured in the logs.

EDIT: Added one more change to this PR:
- If the logger is outputting to a file, makes sure that the directory for that file exists (otherwise an error will be thrown)

### DIFF
--- a/src/data_pipeline.py
+++ b/src/data_pipeline.py
@@ -39,33 +39,43 @@ def get_args() -> argparse.Namespace:
     parser.add_argument(
         "--shape_individual_plants",
         help="Assign an hourly profile to each individual plant with EIA-only data, instead of aggregating to the fleet level before shaping.",
-        type=bool,
         default=True,
+        action=argparse.BooleanOptionalAction
     )
     parser.add_argument(
         "--small",
         help="Run on subset of data for quicker testing, outputs to outputs/small and results to results/small.",
-        type=bool,
         default=False,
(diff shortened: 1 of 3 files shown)


miloknowles:
FWIW, the `--small` option only works right now if you also do `--flat` so that EIA-930 cleaning is skipped.

grgmiller:
Thanks, Milo. I'll take a look. I think that maybe we may get rid of the -small argument altogether as part of our upcoming infrastructure improvements, so not sure if it is worth our time to update it further.

miloknowles:
> Thanks, Milo. I'll take a look. I think that maybe we may get rid of the -small argument altogether as part of our upcoming infrastructure improvements, so not sure if it is worth our time to update it further.

I think having a `--small` option is still useful for testing and debugging. Currently, I'm using it as a faster way to test out a CI/CD pipeline for OGE.

grgmiller (approved):
Looks good as far as I can tell.

PR #289, opened 2023-02-25
Small logging updates

grgmiller:
Currently, the pipeline outputs the log file to `outputs_folder("data_pipeline.log")`, but it would be better if the log were saved to the directory for the specific year for which the pipeline was run, e.g. `outputs_folder("{year}/data_pipeline.log")`. However, the root logger is configured outside of `main()` where the year was defined. To get around this, I just made the final step of the data pipeline `shutil.move(outputs_folder("data_pipeline.log"), outputs_folder(f"{year}/data_pipeline.log"))`. However, I'm not sure if this is a good way to do this. @miloknowles would there be a better / easy way to write directly to the specific year folder?

This PR also makes a few small formatting changes:
- formats `logging_util.py` using black
- Removes the indent from all logging messages

### DIFF
--- a/src/data_pipeline.py
+++ b/src/data_pipeline.py
@@ -24,11 +24,6 @@ from filepaths import downloads_folder, outputs_folder, results_folder
 from logging_util import get_logger, configure_root_logger
 
 
-# Log the print statements to a file for debugging.
-configure_root_logger(logfile=outputs_folder("data_pipeline.log"))
-logger = get_logger("data_pipeline")
-
-
 def get_args() -> argparse.Namespace:
     """Specify arguments here.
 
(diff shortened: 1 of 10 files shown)


miloknowles:
I would just move the `configure_root_logger` call inside of main then, so that you can include the year in the filepath. No harm in doing that.

Btw, the only reason this _wouldn't_ work is if `configure_root_logger` was called somewhere else first with a different filepath (e.g in one of the imported modules). That isn't the case currently, but I would remove the file `src/__init__.py` just to be safe.

src/data_pipeline.py
```diff
@@ -550,6 +550,11 @@ def main():
     hourly_consumed_calc.run()
     hourly_consumed_calc.output_results()
 
+    # move the log file into the specific year output folder
+    shutil.move(
+        outputs_folder("data_pipeline.log"), outputs_folder(f"{year}/data_pipeline.log")
+    )
```
miloknowles, line 556: This will work, but it's probably easier to move the `configure_root_logger` call inside of `main` as you suggested.

miloknowles (approved):

grgmiller:
Hi @miloknowles I updated the location of the logger configuration to be under `main()` in `data_pipeline`. Could you please make a look and make sure this looks ok?

miloknowles (approved):
Looks good! I'd make sure to run the pipeline briefly to make sure it works, but it's good as far as I can see.

PR #290, opened 2023-02-28
Fill missing generated rates

grgmiller:
When outputting power sector data with consumed rates using the new warnings, the warnings caught that there were some missing generated rates in the outputs. After looking into this, it looks like this was happening because we were implementing `.fillna(0)` before we replaced `inf` values with `na`. 

This PR moves the fillna after the `inf` replacement, and also explicitly only fills na values that were a result of `net_generation_mwh` (the denominator in the generated rate calculation) being zero. This means that the expected behavior should be:
- If there is zero net generation but positive emissions mass, the reported emission rate should be zero instead of NA.
- If there is some other reason why the generated rate is missing, these missing values should be left so that they are flagged by our validation checks, since this is unexpected behavior.

### DIFF
--- a/src/output_data.py
+++ b/src/output_data.py
@@ -275,12 +275,9 @@ def write_generated_averages(ba_fuel_data, year, path_prefix, skip_outputs):
                         avg_fuel_type_production[f"{emission}_mass_lb{emission_type}"]
                         / avg_fuel_type_production["net_generation_mwh"]
                     )
-                    .fillna(0)
                     .replace(np.inf, np.NaN)
                     .replace(-np.inf, np.NaN)
-                    .replace(
-                        np.NaN, 0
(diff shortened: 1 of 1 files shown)


src/output_data.py
```diff
@@ -275,12 +275,9 @@ def write_generated_averages(ba_fuel_data, year, path_prefix, skip_outputs):
                         avg_fuel_type_production[f"{emission}_mass_lb{emission_type}"]
                         / avg_fuel_type_production["net_generation_mwh"]
                     )
-                    .fillna(0)
                     .replace(np.inf, np.NaN)
                     .replace(-np.inf, np.NaN)
-                    .replace(
-                        np.NaN, 0
-                    )  # TODO: temporary placeholder while solar is broken. Eventually there should be no NaNs.
```
grgmiller, line 283: This specific change should have no functional impact, this was just duplicative

gailin-p (approved):

PR #291, opened 2023-03-01
Change logfile location to data quality metrics results

grgmiller:
For transparency, I wanted to propose that we change the location of the data pipeline log file from our outputs folder (which gets archived on Zenodo) to the `data_quality_metrics` directory, which users can download directly from the data download page.  This will help give data users transparency into the options, steps, and warnings associated with the data pipeline, if they choose to take a look at this. 

I also realized that we were specifying the location of the logfile before we created all of the sub directories, so I moved the logfile configuration to after the initial code that creates the directory structure.

### DIFF
--- a/src/data_pipeline.py
+++ b/src/data_pipeline.py
@@ -72,6 +72,17 @@ def main():
     args = get_args()
     year = args.year
 
+    # configure the logger
+    # Log the print statements to a file for debugging.
+    configure_root_logger(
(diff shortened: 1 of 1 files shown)


miloknowles:
@grgmiller FYI, the function that configures the logger will create the necessary subdirectories that it needs, so the order shouldn't actually matter

miloknowles (approved):

grgmiller:
Ok thanks! I moved the order back to its original order.

Issue #281, opened 2023-02-15
Validate primary fuel methodology that should be used for assigning a plant's fuel category


grgmiller:
The OGE pipeline currently can use several methods to identify a plant's primary fuel: based on the fuel consumption, based on net generation, or based on nameplate capacity (see [docs](https://docs.singularity.energy/docs/open-grid-emissions-docs/primary_fuel-assigning-primary-fuels)). The plant primary fuel identification is important because that is what is used to categorize each plant into the fuel categories that we use to match plant data up with the hourly EIA-930 data, and is also what we use when we calculate fleet-specific emission rates. Thus, ideally, we should be trying to categorize each plant into the fuel category that is used by EIA-930.

We currently default to identifying the primary fuel based on fuel consumption, but there is a chance that this approach, while technically accurate, may not match up with the fuel category used by EIA-930. For example, if there is a plant with a 100 MW coal generator that runs at 50% capacity and a 90 MW natural gas generator that runs at 90% capacity, our pipeline would likely identify this plant as a natural gas generator, because it is likely consuming more natural gas fuel (by mmbtu) than it is coal. However, just looking at the nameplate capacity of the plant, it has more coal generating capacity than gas capacity, so it might be categorized as a coal plant by the ISO and thus in EIA-930 (even if in a specific year, it did burn more gas than coal). 

We have never validated whether one primary fuel identification method generally performs better than another. There are a few ways we could consider trying to validate which method we should use: We should try grouping plants into fuel categories based on different primary fuel identification methods, then:
1. Compare the error between total reported net generation for that fuel category in EIA-930 and total net generation in that category from our data.
2. Compare the error between total emissions calculated in OGE to total emissions calculated using the fleet emissions factors multiplied by EIA-930 net generation. 

We should also generally look at how much each of these different primary fuel identification methods actually impacts the categorization of plants - how many plants does this affect? Are these plants large or small?

Also, we should start exporting the plant_primary_fuel table as part of our intermediate outputs.

Issue #260, opened 2022-11-27
Updating manual tables


grgmiller:
The `data/manual` folder contains several manual tables. This issue describes how each of these are updated

- `ba_reference`: update using `manually_update_ba_reference.ipynb`. Could be updated on a quarterly basis
- `emission_factors_for_co2_ch4_n2o`: triggered by updates to EPA AP-42 or IPCC
- `emission_factors_for_nox`: triggered by updates to EPA AP-42 or IPCC
- `emission_factors_for_so2`: triggered by updates to EPA AP-42 or IPCC
- `energy_source_groups`: check fuel categories used by EIA-930 and eGRID each year. Also check all fuel codes in the EIA-923 and EIA-860 tables. Check https://catalystcoop-pudl.readthedocs.io/en/latest/data_dictionaries/codes_and_labels.html#energy-sources-eia
- `epa_eia_crosswalk_manual`: run `manually_identify_crosswalk_updates`. Check updates to PSDC. 
- `geothermal_emission_factors`: Triggered by updates to AP-42 or IPCC
- `ipcc_gwp`: Triggered by release of new IPCC Assessment Report every few years
- `physical_ba`: check `transmission_distribution_owner_name` each year
- `plants_not_connected_to_grid`: Check PSDC and eGRID
- `steam_units_to_remove`: Check if there are any plant_id_epa in CEMS that do not map to EIA and only report steam
- `updated_oth_energy_source_codes`: run `manually_update_OTH_fuel_code` and check for plants with OTH fuel code
- `utility_name_ba_code_map`: check for utility names that are not mapped to a ba code

grgmiller:
2022 Update:

- [x] `ba_reference`: no update to the FERC table, new retirements of GLHB and GRIF according to EIA

- [x] `default_gross_to_net_ratios.csv`: No updates
- [ ] `eGRID2020_crosswalk_of_EIA_ID_to_EPA_ID.csv`: 
- [ ] `emission_factors_for_co2_ch4_n2o`: No updates to AP-42 or IPCC
- [x] `emission_factors_for_nox`: No updates to AP-42 or IPCC
- [x] `emission_factors_for_so2`: No updates to AP-42 or IPCC
- [ ] `energy_source_groups`: 
- [ ] `epa_eia_crosswalk_manual`: 
- [ ] `geothermal_emission_factors`: 
- [ ] `ipcc_gwp`: 
- [ ] `physical_ba`:
[...]

Issue #216, opened 2022-09-04
Convert CEMS steam load to gross generation


grgmiller:
A more careful reading of the EPA's Power Sector Emissions Data Guide revealed the following about the reported values for `steam_load_klb`, which we were previously unsure how to interpret (https://github.com/singularity-energy/open-grid-emissions/issues/103):

> Certain EGUs are combined cycle combustion turbines, which generate electricity from a
traditional gas turbine as well as a steam turbine that is powered by the heat of the gas turbine.
These units are not required to report electrical generation from the steam turbine (although
many combined cycle EGUs do include the generation from the steam turbine in their electrical
generation data).
> EGUs that generate steam **have the option to report steam load in thousand pounds of steam
per hour (1000lb/hr) instead of electrical generation in MWh**.

Thus, it seems that we may want to look into converting steam load to gross generation when a unit only reports steam load and no gross generation.

Questions for further research:
- Do any units report both gross generation and steam load for the same hour?
- Is there a standard conversion factor between steam load and gross generation?
- Are there any unit-specific datapoints available about the steam quality/pressure that would result in a more accurate conversion?

grgmiller:
This may require a refresher on my basic thermodynamic equations. I think if we know the steam flow rate in klb/hr, we need the enthalpy of the steam to convert to MW. I'm not sure if there is data available about the enthalpy of the steam at each plant, but we could assume a typical temperature and pressure. 

Resources:
- https://www5.eere.energy.gov/manufacturing/tech_deployment/amo_steam_tool/equipTurbine?random=Outlet
- https://www.govinfo.gov/content/pkg/CFR-2013-title40-vol7/pdf/CFR-2013-title40-vol7-sec60-4350.pdf

grgmiller:
Also see reporting instructions for more information about steam load data: https://www.epa.gov/system/files/documents/2024-03/ecmps_emissions_reporting_instructions_0.pdf, found at https://www.epa.gov/power-sector/ecmps-reporting-instructions.

It looks like the conversion from klb steam to MWh is approximately 0.3-0.35 MWh per klb, based on https://energystar.my.site.com/PortfolioManager/s/article/How-do-I-enter-District-Steam-that-has-been-billed-in-pounds-1600088536249 and https://www.abraxasenergy.com/energy-resources/toolbox/conversion-calculators/power/

Issue #134, opened 2022-07-06
Integrate CEMS data for units that only report NOx


grgmiller:
A handful of units that report to CEMS only report NOx emissions data, but report zero data for gross load, fuel consumption, and other pollutants. As of https://github.com/singularity-energy/hourly-egrid/pull/133, we now remove these observations from the dataset in favor of using EIA-923 data.

However, in the future, it may be worth adding this standalone hourly NOx data back into the results, as it could help us shape the EIA data.

PR #322, opened 2023-12-09
Update to 2022 data

grgmiller:
### Purpose
This PR updates the OGE pipeline to work with 2022 data, and also updates the manual tables.

### What the code is doing
Updates default years to 2022 (Fixes CAR-3399)
Updates source of eGRID2020 data to the v2 file (not used in the pipeline, only for comparison)

Updates reference tables (see https://github.com/singularity-energy/open-grid-emissions/issues/260) (Fixes CAR-3349)
- [x] `ba_reference`: no update to the FERC table, new retirements of GLHB and GRIF according to EIA
- [x] `default_gross_to_net_ratios.csv`: No updates
- [x] `eGRID2020_crosswalk_of_EIA_ID_to_EPA_ID.csv`: updated based on eGRID2021. One new plant added to list
- [x] `emission_factors_for_co2_ch4_n2o`: No updates to AP-42 or IPCC
- [x] `emission_factors_for_nox`: No updates to AP-42 or IPCC. Added several new factors for boiler configurations that were not previously added, but are in the 2022 data. 
- [x] `emission_factors_for_so2`: No updates to AP-42 or IPCC, Added several new factors for boiler configurations that were not previously added, but are in the 2022 data. 
- [x] `energy_source_groups`: no changes based on pudl metadata
- [x] `epa_eia_crosswalk_manual`: used notebook to identify new additions to table
- [x] `geothermal_emission_factors`: no changes to source data
- [x] `ipcc_gwp`: most current report is still AR6
- [ ] `physical_ba`
- [x] `plants_not_connected_to_grid`: no changes based on eGRID2021
- [x] `steam_units_to_remove`: Not updated since steam units no longer being removed.
- [x] `updated_oth_energy_source_codes`: ran notebook, no new matches needed
- [x] `utility_name_ba_code_map`: ran notebook, added several new maps, sorted alphabetically.

`load_data.py`
- Although much of the EPA plant ID to EIA plant mapping has been integrated into the pudl version of the CEMS data, our manual plant mapping from `epa_eia_crosswalk_manual` would not be reflected. Now, whenever cems data is being loaded, we run `update_epa_to_eia_map()` to update the `plant_id_eia` codes
- When loading the EPA-EIA crosswalk, we now load in the manual plant map from eGRID to make sure this is considered in the mapping.

`data_cleaning.py`
- No longer remove steam units: Previously, we had been dropping from the CEMS data any units that only reported steam load and no gross generation (see #103 for context), mostly just to save memory since we were not including this steam data in the outputs. However, based on our updated understanding in #216 we do not necessarily want to be dropping this data. Even though we are not yet implementing this steam to generation conversion, we would have had to have updated the `steam_units_to_remove` manual table, which didn't seem worth it if we are not going to be dropping these units in the future. 
- Drops negative fuel consumption: Plant 10613 reports negative fuel consumption for its WDS fuel in May 2022. This is an error in the input EIA-923 data but our pipeline was not previously set up to handle such cases. For now, we replace these negative fuel consumption values with missing values, and print a warning in the logger.
- Switches back to running `pudl.analysis.allocate_gen_fuel_by_generator_energy_source()` instead of loading the table from pudl. This allows us to see data quality warnings generated during running that part of the pipeline which will help with data quality checks. 
- Deletes the code that dropped all EIA-923 that was missing or zero. I cannot remember why I implemented this, but this was resulting in a lot of missing records in the data outputs. Even if fuel consumption and generation is zero for a plant in the month, that is still useful information and we should not drop it. Note: some data is actually missing in the input EIA-923 data.
- Fixes a bug when identifying partial cems plants. We did not want to assign the "partial cems" methodology to a subplant if the primary fuel of the plant was not fossil based, since that could mean that we were assigning data from a fossil backup generator to shape nuclear generation, for example. This should have been identified using the "plant_primary_fuel" column but it was mistakenly using the "energy_source_code" column, which is the generator-specific energy source. This was leading to a mixed method error at line 1765.

`eia930.py`
- Adds an end date to the TEPC timestamp issue in EIA-930 (Fixes CAR-2506)

`emissions.py`
- For some reason in 2022, none of the JF-fueled plants reported fuel sulfur content data. This means that several plants were missing SO2 emission factors. To fix this, I added a further backstop to the existing `annual_avg_fuel_sulfur_content`. Now, in the case that there is no sulfur content data available for a fuel in a year, the pipeline will check the reported sulfur contents from the previous year to see if it can fill in the annual average value. This means for the 2022 pipeline, the JF generators without a specified sulfur content will use the average JF sulfur content from 2021 (looking at multiple years, this sulfur content does not seem to change from year to year so this seems like a reasonable backstop). In implementing this fix, I split an existing function into two components.
- When checking for missing nox and so2 emission factors, we now only raise this error if these are missing for PM-fuels where there is non-zero fuel consumption. Previously we had been checking in case of missing fuel consumption, but we do not need emission factors for fuel-PMs with zero fuel consumption.

`validation.py`
- now print a sample of the bad data whenever the `test_for_negative_values` check does not pass
- Fixes bug where the incorrect variable was being referenced when testing emissions adjustments
- renames the `check_for_complete_timeseries` check to `check_for_complete_hourly_timeseries`, and adds a `check_for_complete_monthly_timeseries` check. This new check ensures that monthly-resolution data contains all 12 months of data. 

### Testing
Running the pipeline for 2022 (not yet complete)

### Usage Example/Visuals

### Review estimate

### Future work
The following warnings were raised when running the pipeline:
- "oge.oge.validation:249 There are 259 subplants that only contain one part of a combined cycle system. Subplants that represent combined cycle generation should contain both CA and CT parts.": This error is the same as previous years, this is something on our backlog that we haven't gotten around to fixing yet.
- "oge.oge.validation:78 Allocated EIA-923 doesn't match input data for plants": This is a high priority thing to fix but it requires messing with the pudl code. See https://github.com/catalyst-cooperative/pudl/issues/3165
- "oge.oge.validation:140 There are 173 plants where the assigned primary fuel doesn't match the capacity-based primary fuel. It is possible that these plants will categorized as a different fuel in EIA-930": This is expected/same as previous years. Nothing to "fix" here but may influence some future design choices about how primary fuels are assigned. See https://github.com/singularity-energy/open-grid-emissions/issues/281
- We get our expected warnings about missing Nox and SO2 emission factors for fuel cell (FC) prime movers. This is expected. See https://github.com/singularity-energy/open-grid-emissions/issues/70
- "oge.oge.validation:334 Some data being removed has non-zero data associated with it:" We drop cems data where there is zero generation or fuel consumption data. However, some of these records have non-zero steam load and nox mass. This is a known issue and is tied to https://github.com/singularity-energy/open-grid-emissions/issues/134
- "oge.oge.validation:374 There are 212 subplants at 142 plants for which there is zero gross generation associated with positive net generation.", "oge.oge.validation:411 There are 53 subplants at 41 plants for which there is zero net generation associated with positive gross generation.", "oge.oge.validation:450 The following plants have annual net generation that is >125% of annual gross generation:" This is raised when EIA data is linked to CEMS data, and the generation numbers are inconsistent between the two sources. This happens with all years of data and is not unique to 2022. This may be due to inconsistent reporting between sources, or potentially bad crosswalking. This will require future investigation in our backlog of issues. 
- [ ] "oge.oge.gross_to_net_generation:93 The following subplants are missing default GTN ratios. Using a default value of 0.97": For some reason these plants do not have prime-mover related default factors from EIA. We should  look into the prime mover code of these plants. 
- "oge.oge.validation:1606 Potentially anomalous co2 factors detected for the following plants" - there are some plants that have abnormally high or low EFs. These need to be manually looked into to understand the source of these anomalies. Some might just be input data quality issues, or there is an issue with our processing pipeline, or they are just abnormal plants and the data is accurate. 

### Checklist


src/oge/data_cleaning.py
```diff
@@ -1002,7 +1018,8 @@ def clean_cems(year: int, small: bool, primary_fuel_table, subplant_emission_fac
     )
 
     # manually remove steam-only units
-    cems = manually_remove_steam_units(cems)
+    # NOTE(greg): disabling this for the 2022 data release
+    # cems = manually_remove_steam_units(cems)
```
rouille, line 1022: Why do we stop removing these units in 2022?

src/oge/consumed.py
```diff
@@ -438,13 +441,13 @@ def _load_rates(self):
                         axis=1,
                     )
 
-                # Cut off emissions at 9 hours after UTC year
-                emissions = emissions[: f"{self.year+1}-01-01 09:00:00+00:00"]
+                # Cut off emissions at 8 hours after UTC year
+                emissions = emissions[: f"{self.year+1}-01-01 08:00:00+00:00"]
```
rouille, line 445: How does it work? I thought the emission would be hourly over a full year in UTC time. Are we doing this in Pacific Time instead. And why was it nine hours before, did we have Alaska in 2021 and not 2022?

src/oge/consumed.py
```diff
@@ -438,13 +441,13 @@ def _load_rates(self):
...
                 rates[(adj, pol)] = emissions
 
         # Make generation data frame
         generation = pd.DataFrame(data=gens)
-        generation = generation[: f"{self.year+1}-01-01 09:00:00+00:00"]
+        generation = generation[: f"{self.year+1}-01-01 08:00:00+00:00"]
```
rouille, line 450: same question

src/oge/data_cleaning.py
grgmiller: Sorry this is unclear! This is not just for 2022, we are stopping using this method altogether. See note about this in the main PR description. I'll delete this whole line.

src/oge/consumed.py
grgmiller: So this is just saying that we only want to do the calculations through midnight pacific time on 12/31, which would be 8am UTC time on Jan 1. Not sure why this was previously set to 9 hours, since we do not do the consumed calculations for AK or HI. I'm also not sure why this didn't raise a key error in previous years, but it does now, since 2023-01-01 9:00 does not exist in the data.

src/oge/consumed.py
grgmiller: See above

src/oge/consumed.py
rouille: Should we also set the lower bound? So, it starts on 2022-01-01 Pacific Time, i.e,  do:
```
emissions = emissions[f"{self.year}-01-01 08:00:00+00:00": f"{self.year+1}-01-01 08:00:00+00:00"]
```
Right now, we probably have the last 8 hours in 2021/12/31.

src/oge/data_cleaning.py
rouille: I understand it will apply to all years from now on. My question is why do we stop removing steam-only units. Do we feel confident estimating emissions from these generators moving forward?

grgmiller:
Updates since last review: (in general, these are aimed at reducing the number of missing values in the output data)

`data_cleaning.py`
- Adds a new function to help with validating missing data later in the pipeline `inventory_input_data_sources`. When we identify missing months of data, we want to know whether we introduced that missing data, or if the data was already missing in the input data sources. This function identifies, for each plant-month, whether there was any data in CEMS or the EIA-923 generation and fuel table (the two definitive sources of generation and fuel consumption data) for that plant-month. It also indicates whether there was any non-zero input data for that plant month. This table is outputted to the outputs folder. 
- We had previously been removing data from the CEMS table if there were all zeros reported for a specific unit month. However, all zeros are still data, and by removing this, we were introducing missing data further down in the pipeline. We stop using `remove_cems_with_zero_monthly_data()`.
- Sometimes in CEMS, there is a month with less than a single day of data (and typically only a single hour of data). The function that removed these unit months was previously removing data with fewer than 600 monthly observations, but the intent was only to remove data that is mostly incomplete except for a few hours. We fix this bug. 
- Fixes a bug when identifying partial cems subplatnts. We only want to flag the subplant as having partial data if both the data from CEMS and EIA is non-zero.
- Fixes a bug when identifying balancing areas where some plants were not being assigned a BA. We now use the most recent BA assignment avaialble, even if there is a missing BA assignment for the current year.

`output_data.py`
- When outputting plant metadata, we fix an issue where incorrect metadata for shaped EIA plants was getting outputted, and re-factored the code a bit to make it easier to follow. 

`validation.py`
- We update the check for complete monthly timeseries to make use of the input data inventory. This check now flags whether there is less than 12 months of output data for each grouping, as well as where there are more missing data values than expected based on missing inputs.

rouille (approved):
Thanks

PR #337, opened 2024-01-30
Add eGRID2022 to file downloads

grgmiller:
### Purpose
Adds 2022 eGRID data to the list of eGRID downloads. This is needed to support the addition of 2022 EFs for the Grid Carbon API. Closes CAR-3729

### What the code is doing
In the process of adding the 2022 data, I realized that the naming conventions for the historical eGRID data is not consistent, which would have led to errors when trying to load egrid data in some of the validation and visualization functions. I updated the download_egrid_files function to standardize the egrid file name format across years. as part of this, I moved the list of files from `pipeline` to `download_data`.

### Testing
I ran `download_data.download_egrid_files()`. All eGRID files (2018-2022) were downloaded successfully and have a consistent file name.

### Review estimate
5 min

### Future work
We do not currently use the eGRID data in the main part of the pipeline, but at some point we should compare our 2022 data to the eGRID 2022 data. 

### Checklist


grgmiller:
Converting this back to a draft - realizing that there is more work to do in order to pull in the 2022 data

PR #339, opened 2024-02-03
Fix circular import of CLEAN_FUELS and add constants

grgmiller:
### Purpose
Some of the changes implemented in https://github.com/singularity-energy/open-grid-emissions/pull/337 introduced a circular import between the `load_data` and `emissions` modules, due to the import of `CLEAN_FUELS` into the `load_data` module. 

### What the code is doing
Moves `CLEAN_FUELS` (along with some other constants) to a new `constants.py` module to prevent a circular import. 

While in the process of fixing this specific circular import, I also searched the repo for where other constants / assumptions were specified and consolidated them into `constants`.

In `emissions.calculate_electric_allocation_factor()` updated the calculation steps and column names to match the process described in the "Adjusting Emissions for CHP" docs (there is no methodological change here, just separating out several steps and renaming some variables to be more clear). 

### Testing
Tested importing OGE with this new branch in another package, and the circular import error was resolved. 

### Where to look
All files

### Usage Example/Visuals
N/A

### Review estimate
5 min

### Future work
N/A

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,7 @@ There are many ways that you can contribute!
 ## Repository Structure
 ### Modules
 - `column_checks`: functions that check that all data outputs have the correct column names
+- `constants`: specifies conversion factors and constants used across all modules
 - `data_pipeline`: main script for running the data pipeline from start to finish
 - `download_data`: functions that download data from the internet
 - `data_cleaning`: functions that clean loaded data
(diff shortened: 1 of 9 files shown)


rouille (approved):
Cleaner. Thanks

PR #344, opened 2024-02-14
Updates python version in pyproject.toml and removes requirements.txt

Rdbaker:
### Purpose

This change:
* updates the versions of the python requirement to align with PUDL, so that this module can be installed
* removes the unused `requirements.txt` file, since we have 3 other dependency management files in this repo (correct me if we want that, but it wasn't referenced in the README at all)

### What the code is doing

updating the python version in pyproject.toml

### Testing

I will test this change by trying to install it in Greta

### Where to look

n/a

### Usage Example/Visuals

n/a

### Review estimate

5 minutes

### Future work

none for this change

### Checklist


Rdbaker:
tested that it was able to be installed in Greta

grgmiller:
Right now the target branch is `main` - if we test integration using this branch and everything works, we can merge directly into main and do a `v0.3.2` release of OGE; otherwise, we probably want to first merge into `development` in case there are any other patch changes we need to make during the testing process for integration.

grgmiller:
If we update the version, this also needs to be updated in the `CITATION.cff` file along with the release date.

Rdbaker:
> If we update the version, this also needs to be updated in the `CITATION.cff` file along with the release date.

why are we using both CITATION.cff and pyproject.toml to specify the version?

grgmiller:
> why are we using both CITATION.cff and pyproject.toml to specify the version?

I think they serve different purposes. the citation file is to help with the citability for academic purposes, so the version number is specified there as well. https://citation-file-format.github.io/

pyproject.toml
```diff
@@ -33,7 +33,6 @@ dependencies = [
   "statsmodels",
   "coloredlogs",
   "s3fs",
-  "catalystcoop-pudl@git+https://github.com/singularity-energy/pudl.git@oge_release",
```
grgmiller, line 36: We cannot remove PUDL from the dependencies of OGE

grgmiller (changes requested):
We

grgmiller (approved):
Looks good

PR #345, opened 2024-02-22
Update conda environment

grgmiller:
### Purpose
In our most recent releases, we updated the pipenv and readme instructions for using pipenv, but we have not kept the conda environment up to date since we were not actively using it. However, a user was trying to run OGE using conda, and was running into issues when trying to install the conda environment. This PR fixes those issues.

### What the code is doing
N/A

### Testing
Installed a fresh conda environment, and ran the full OGE pipeline for 2019 with the `--skip_outputs` option on a windows machine.

### Where to look
Modifies `environment.yml` and the `README`

### Usage Example/Visuals
N/A

### Review estimate
5 min to look at code, longer if testing on Mac OS.

### Future work
N/A

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,17 +10,9 @@ Please check out [our documentation](https://docs.singularity.energy/docs/open-g
 The Open Grid Emissions Dataset can be [downloaded here](https://singularity.energy/open-grid-emissions). An archive of previous versions of the dataset and intermediate data outputs (for research and validation purposes) can be found on [Zenodo](https://zenodo.org/communities/singularity-energy?page=1&size=20).
 
 ## Installing and running the data pipeline
-A couple of installation methods are available. First, navigate to the folder where you want to save the repository and run the following commands:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -133,11 +138,17 @@ Open anaconda prompt, use `cd` to navigate to the directory where your local fil
...
 ```
-and activate the environment:
+Installation requires that the conda channel-priority be set to "flexible". This is the default behavior, so if you've never manually changed this, you shouldn't have to worry about this. However, if you receive an error message like "Found conflicts!" when trying to install the environment, try setting your channel priority to flexible by running the following command:`conda config --set channel_priority flexible` and then re-running the above commands.
+
+The final step is to install the `oge` package itself in the conda environment. To do so, run:
 ```bash
 conda activate open_grid_emissions
+pip install build
+python -m build
+pip install .
```
rouille, line 148: I guess for the development setup we should recommend doing `pip install --editable .`. This is what we suggest doing below for `pipenv`

grgmiller: Good catch! I'll update this.

PR #346, opened 2024-02-23
Fix filepaths for s3

grgmiller:
### Purpose
When importing OGE into another module on windows, and attempting to load data from s3, the current way filepaths are handled results in a FileNotFound error.

For example, running `oge.filepaths.results_folder("2022/plant_data/plant_static_attributes.csv")` returns:
```
's3://open-grid-emissions/open_grid_emissions_data\\results\\2022/plant_data/plant_static_attributes.csv'
```
Where `os.path.join()` is concatting parts of the filepath together, it is adding a "\\" rather than a "/".
When attempting to read from this location, for example using `pd.read_csv(oge.filepaths.results_folder("2022/plant_data/plant_static_attributes.csv")`, this results in:
```
FileNotFoundError: open-grid-emissions/open_grid_emissions_data\results\2022/plant_data/plant_static_attributes.csv
```
Here, the mix of forward and back slashes does not seem to be compatible with reading from s3. 
If instead I try `pd.read_csv(oge.filepaths.results_folder("2022/plant_data/plant_static_attributes.csv").replace("\\","/"))`, this successfully returns the dataframe

### What the code is doing
This PR adds `.replace("\\","/")` to all of the functions that return folder locations where data is saved to help prevent this issue in any package that imports OGE and reads the data from s3.

### Testing
Did a fresh install of the local oge version, and ran the pipeline long enough to check that local data loading still works.

### Review estimate
Code review: < 5 min. If you test on a mac, it may take longer. 

### Future work

### Checklist

### DIFF
--- a/src/oge/filepaths.py
+++ b/src/oge/filepaths.py
@@ -23,24 +23,24 @@ def top_folder(rel=""):
 
 
 def reference_table_folder(rel=""):
-    return os.path.join(top_folder("reference_tables"), rel).replace("\\","/")
+    return os.path.join(top_folder("reference_tables"), rel).replace("\\", "/")
 
 
 def data_folder(rel=""):
     """Returns a path relative to the `data` folder."""
-    return os.path.join(get_data_store(), rel).replace("\\","/")
+    return os.path.join(get_data_store(), rel).replace("\\", "/")
 
 
 def downloads_folder(rel=""):
-    return os.path.join(data_folder("downloads"), rel).replace("\\","/")
+    return os.path.join(data_folder("downloads"), rel).replace("\\", "/")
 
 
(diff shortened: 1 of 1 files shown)


rouille:
I am worried that by fixing the path for s3 we break the path for Windows' users running locally

grgmiller:
> I am worried that by fixing the path for s3 we break the path for Windows' users running locally

I did test this locally on my windows laptop and this still works to load the data. I've never had an issue using forward slashes for filepaths on windows

>Testing
Did a fresh install of the local oge version, and ran the pipeline long enough to check that local data loading still works.

rouille (approved):
Thanks

PR #352, opened 2024-03-19
Access a specific GWP when calculating CO2 mass equivalent value

rouille:
### Purpose
Allow to calculate a CO2 mass equivalent value using a Global Warming Potential (GWP) value published in any available IPCC's assessment report (AR).

### What the code is doing
Modify the `calculate_co2e_mass` function such that its second parameter can be used to fetch not only the most recent AR that was available in the given year as done previously but also to directly fetch the AR.

### Testing
Tested manually

### Where to look
In the `oge.emissions` module

### Usage Example/Visuals
```python
>>> import pandas as pd
>>> from emissions import calculate_co2e_mass
>>> calculate_co2e_mass(pd.DataFrame(), "AR32")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/brdo/Singularity/open-grid-emissions/src/oge/emissions.py", line 418, in calculate_co2e_mass
    raise ValueError(
ValueError: version must be a year in [1995, 2021] or a valid AR
>>> calculate_co2e_mass(pd.DataFrame(), 2022)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/brdo/Singularity/open-grid-emissions/src/oge/emissions.py", line 418, in calculate_co2e_mass
    raise ValueError(
ValueError: version must be a year in [1995, 2021] or a valid AR
```

### Review estimate
2min

### Future work
N/A

### Checklist

### DIFF
--- a/src/oge/emissions.py
+++ b/src/oge/emissions.py
@@ -378,33 +378,50 @@ def adjust_emissions_for_biomass(df):
     return df
 
 
-def calculate_co2e_mass(df, year, gwp_horizon=100, ar5_climate_carbon_feedback=True):
+def calculate_co2e_mass(df, version, gwp_horizon=100, ar5_climate_carbon_feedback=True):
     """Calculate CO2-equivalent emissions from CO2, CH4, and N2O. This is done by
     choosing one of the IPCC's emission factors for CH4 and N2O.
+
+    Note that the `version` parameter controls the version of the IPCC AR that will be
+    chosen. If `version` is an int then it will be interpreted as a year and the most
+    recent AR that was available that year will be picked. If `version` is a str then
+    it will be interpreted as an IPCC Assessment Report (AR). Therefore, valid
+    `version` are: int >= 1995 and any str equals to SAR, TAR, AR4, AR5, AR5_cc, or AR6.
+    Note that when `version` is a str the `ar5_climate_carbon_feedback` has no effect
+    since the user has direct access to AR5 or AR5_cc via the `version` parameter.
+
 
     If the `fuel_consumed_for_electricity_units` column is available, we also compute
     the adjusted emissions.
     """
     df_gwp = load_data.load_ipcc_gwp()
 
-    # use the most recent AR that was available in the given year
-    ipcc_version = df_gwp.loc[df_gwp["year"] <= year, "year"].max()
+    if (
+        isinstance(version, int) and version >= df_gwp["year"].min()
+    ) or version == "latest":
+        year = version if isinstance(version, int) else df_gwp["year"].max()
 
-    # use the most recent AR that was available in the given year
-    most_recent_AR_year = df_gwp.loc[df_gwp["year"] <= year, "year"].max()
+        # use the most recent AR that was available in the given year
+        most_recent_AR_year = df_gwp.loc[df_gwp["year"] <= year, "year"].max()
 
-    # identify the AR
-    ipcc_version = list(
-        df_gwp.loc[df_gwp["year"] == most_recent_AR_year, "ipcc_version"].unique()
-    )
-    if len(ipcc_version) > 1:
-        if "AR5" in ipcc_version and ar5_climate_carbon_feedback:
-            ipcc_version = "AR5_cc"
-        elif "AR5" in ipcc_version and not ar5_climate_carbon_feedback:
-            ipcc_version = "AR5_cc"
+        # identify the AR
+        ipcc_version = list(
+            df_gwp.loc[df_gwp["year"] == most_recent_AR_year, "ipcc_version"].unique()
+        )
+        if len(ipcc_version) > 1:
+            if "AR5" in ipcc_version and ar5_climate_carbon_feedback:
+                ipcc_version = "AR5_cc"
+            elif "AR5" in ipcc_version and not ar5_climate_carbon_feedback:
+                ipcc_version = "AR5_cc"
+        else:
+            ipcc_version = ipcc_version[0]
+    elif version in df_gwp["ipcc_version"].unique():
+        ipcc_version = version
     else:
-        ipcc_version = ipcc_version[0]
-
+        raise ValueError(
+            f"version must be a year >= {df_gwp['year'].min()} or a valid AR"
+        )
+    print(ipcc_version)
     gwp_to_use = df_gwp[df_gwp.ipcc_version == ipcc_version]
 
     ch4_gwp = gwp_to_use.loc[


src/oge/emissions.py
```diff
@@ -378,32 +378,47 @@ def adjust_emissions_for_biomass(df):
...
+    interpreted as the AR.
+
     If the `fuel_consumed_for_electricity_units` column is available, we also compute
     the adjusted emissions.
     """
     df_gwp = load_data.load_ipcc_gwp()
 
-    # use the most recent AR that was available in the given year
-    ipcc_version = df_gwp.loc[df_gwp["year"] <= year, "year"].max()
+    if isinstance(version, int) and version in df_gwp["year"].unique():
```
grgmiller: So the year here does not need to be explicitly in the list of years - for example, if we enter version = 2022, then we'd want this to return 2021 as the most recent AR year. I think we could just remove the second test entirely and just check if it is an int.

src/oge/emissions.py
```diff
@@ -378,32 +378,47 @@ def adjust_emissions_for_biomass(df):
...
+        year = version
+        # use the most recent AR that was available in the given year
+        ipcc_version = df_gwp.loc[df_gwp["year"] <= year, "year"].max()
```
grgmiller, line 399: This seems to be a duplicate of line 402. I think we could just keep 402

src/oge/emissions.py
```diff
@@ -378,32 +378,47 @@ def adjust_emissions_for_biomass(df):
...
+            df_gwp.loc[df_gwp["year"] == most_recent_AR_year, "ipcc_version"].unique()
+        )
+        if len(ipcc_version) > 1:
+            if "AR5" in ipcc_version and ar5_climate_carbon_feedback:
+                ipcc_version = "AR5_cc"
+            elif "AR5" in ipcc_version and not ar5_climate_carbon_feedback:
+                ipcc_version = "AR5_cc"
+        else:
+            ipcc_version = ipcc_version[0]
+    elif isinstance(version, str) and version in df_gwp["ipcc_version"].unique():
```
grgmiller: Here, if someone just enters "AR5" and sets ar5_climate_carbon_feedback=True, this will not pass the second test.

src/oge/emissions.py
```diff
@@ -378,32 +378,47 @@ def adjust_emissions_for_biomass(df):
     return df
 
 
-def calculate_co2e_mass(df, year, gwp_horizon=100, ar5_climate_carbon_feedback=True):
+def calculate_co2e_mass(df, version, gwp_horizon=100, ar5_climate_carbon_feedback=True):
```
grgmiller, line 381: if we are changing the name of this variable, we will need to go through the repo and make sure that this variable isn't explicitly named anywhere, otherwise we will need to change it. Likewise with other repos that import OGE and may use this function.

grgmiller (commented):
Some changes requested. It would be helpful if you could test passing the following to see what it returns:
- 2022 (should return AR6 numbers)
- "AR4" (should return AR4)
- Test several variations of AR5 with and without the CC feedback.

src/oge/emissions.py
rouille: Not sure I understand. This is a positional argument, the name of the variable you pass does not matter.

src/oge/emissions.py
rouille: What about years < 1995, that will raise an error. Are we fine we that?

src/oge/emissions.py
rouille: The `ar5_climate_carbon_feedback` is not used when `version` is a string. If you pass `AR5` then `ipcc_version` is set to `AR5`, if you pass `AR5_cc` then `ipcc_version` is set to `AR5_cc`

rouille:
> Some changes requested. It would be helpful if you could test passing the following to see what it returns:
> 
>     * 2022 (should return AR6 numbers)
> 
>     * "AR4" (should return AR4)
> 
>     * Test several variations of AR5 with and without the CC feedback.

In the last version of the code where I have a print of the `ipcc_version` (to be removed before merging) at the end of the `if`/`else` logic, I have got the following:
```
>>> import pandas as pd
>>> from emissions import calculate_co2e_mass
>>> calculate_co2e_mass(pd.DataFrame(), 2022)
AR6
>>> calculate_co2e_mass(pd.DataFrame(), "latest")
AR6
>>> calculate_co2e_mass(pd.DataFrame(), "AR4")
AR4
>>> calculate_co2e_mass(pd.DataFrame(), "AR5")
AR5
>>> calculate_co2e_mass(pd.DataFrame(), "AR5_cc")
AR5_cc
>>> calculate_co2e_mass(pd.DataFrame(), "AR5", ar5_climate_carbon_feedback=True)
AR5
>>> calculate_co2e_mass(pd.DataFrame(), "AR5", ar5_climate_carbon_feedback=False)
AR5
>>> calculate_co2e_mass(pd.DataFrame(), "AR5_cc", ar5_climate_carbon_feedback=False)
AR5_cc
>>> calculate_co2e_mass(pd.DataFrame(), "AR5_cc", ar5_climate_carbon_feedback=True)
AR5_cc
>>> calculate_co2e_mass(pd.DataFrame(), "AR7")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/brdo/Singularity/open-grid-emissions/src/oge/emissions.py", line 418, in calculate_co2e_mass
    raise ValueError(
ValueError: version must be a year > 1995 or a valid AR
>>> calculate_co2e_mass(pd.DataFrame(), 1990)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/brdo/Singularity/open-grid-emissions/src/oge/emissions.py", line 418, in calculate_co2e_mass
    raise ValueError(
ValueError: version must be a year >= 1995 or a valid AR
>>> 
```

src/oge/emissions.py
rouille: In the last version of the code it raises an error if version is an `int` < the min year found in the data frame

src/oge/emissions.py
grgmiller: Right, but I'm not sure whether we ever explicitly called the variable such as df = calculate_co2e_mass(df=df, year=2022, gwp_horizon=20, ar5_climate_carbon_feedback=False). I just looked through the OGE code and it looks like we didn't do this so this should not be an issue.

src/oge/emissions.py
grgmiller: Got it. Can we add this explanation to the docstring so that someone is not expecting to get the AR5_cc value back if they set this to True?

Maybe in the docstring we should also enumerate all valid values of version, eg an int >=1995 or a string in SAR, TAR, AR4, AR5, AR5_cc, or AR6.

grgmiller (approved):
Thanks Ben, a few small additional documentation changes requested, but otherwise good to go.

src/oge/emissions.py
rouille: Ah!. Sorry I understand now. I don't think I ever called a function by explicitly assigning the parameters.

src/oge/emissions.py
rouille: I have updated the docstring. Let me know what you think.

grgmiller:
Looks good to me! Should be ready to merge.

PR #363, opened 2024-05-18
Calculate electric_allocation_factor by subplant

grgmiller:
### Purpose
As I was exploring OGE data, I noticed that in some months, some subplants have an electric_allocation_factor (EAF) of 0 (meaning no fuel went to electricity production), but have net generation > 0. If net generation is greater than zero, then the electric allocation factor should always be greater than zero. Here's an example of the issue:
image

If we use plant 4042, subplant 1 as an example, here's what I found is happening: This subplant has two generators, 1 and 2. In the EIA-923 data, all of the fuel consumption in May is reported for generator 1, while all of the generation data is reported for generator 2. Since the EAF calculation depends on both factors, The EAF for gen 1 is calculated as 0 (which is multiplied by all of the fuel), and the EAF for gen 2 is 1, but is multiplied by 0 fuel.

This suggests that when calculating the electric allocation factor, we should probably be calculating this at the subplant level rather than the generator level, since the fuel and generation may not always be reported for the same generator in a subplant.

### What the code is doing
Updates the code for calculating electric allocation factors, to calculate them for each subplant-month-fuel group, so that we are not dropping electricity-related fuel on accident.

I also added a new helper function `add_subplant_ids_to_df()` that adds subplant ids to a dataframe, since pattern was re-used in a bunch of places.

In `load_data`, I also added a new argument to `add_report_date` that will allow this function to be used in other repos without failing when the data source is s3.

### Testing
If we look at the results for the example plant 4042 subplant 1 above, for the month of May 2021, we see that using the subplant EAF now allocates non-zero electricity-related fuel consumption to the subplant:
image

Although this has a big impact on specific plants, and the distribution of fuel among generators in EIA-923, this does not have a large impact on the total fuel consumption across the country:
image

I successfully ran the entire pipeline for 2018.

### Review estimate
10 minutes

### Future work

### Checklist


grgmiller:
@gailin-p tagging you FYI

src/oge/emissions.py
```diff
@@ -223,8 +223,18 @@ def identify_geothermal_generator_geotype(year):
...
 
-def adjust_fuel_and_emissions_for_CHP(df):
-    """Allocates total emissions for electricity generation."""
+def adjust_fuel_and_emissions_for_CHP(df: pd.DataFrame) -> pd.DataFrame:
+    """Allocates total emissions for electricity generation.
+
+    Args:
+        df (pd.DataFrame): dataframe containing fuel and emissions columns mapped to
+        subplant_id. Must contain fuel_consumed_mmbtu,
+        fuel_consumed_mmbtu_for_electricity, and net_generation_mwh columns
```
rouille, line 232: I believe in Google Style Python docstrings convention, if the description of a parameter spans multiple lines, then following lines should be indented.

src/oge/emissions.py
```diff
@@ -223,8 +223,18 @@ def identify_geothermal_generator_geotype(year):
...
+
+    Returns:
+        pd.DataFrame: the input dataframe with _for_electricity columns added for all
+        fuel and emissions columns
```
rouille, line 236: same comment here about indentation

src/oge/emissions.py
```diff
@@ -286,48 +296,89 @@ def adjust_fuel_and_emissions_for_CHP(df):
...
+    not always assigned to the same generator within a subplant.
 
-    Requires a dataframe with the following columns: net_generation_mwh, fuel_consumed_mmbtu, fuel_consumed_for_electricity_mmbtu
+    Calculates a unique EAF for each subplant-month-fuel.
+
+
+    Args:
+        df (pd.DataFrame): dataframe fuel_consumed_mmbtu,
+        fuel_consumed_mmbtu_for_electricity, and net_generation_mwh columns, mapped to
+        subplant_id
```
rouille, line 314: same comment about indentation.

src/oge/helpers.py
```diff
@@ -331,3 +332,49 @@ def add_plant_local_timezone(df, year):
...
+    """Helper function for adding subplant_id's to a dataframe based on the `plant_part_to_map`.
+
+    Validates that all rows have non-missing subplant_id's assigned.
+
+    Args:
+        df (pd.DataFrame): Dataframe to add subplant_id's to
+        year (int): the data year
+        plant_part_to_map (str): one of the plant parts in the subplant crosswalk that
+        exists in df to use to map the subplant ids. Acceptable options include:
+        "generator_id", "boiler_id", or "emissions_unit_id_epa"
```
rouille, line 353: same comment about indentation

src/oge/helpers.py
```diff
@@ -331,3 +332,49 @@ def add_plant_local_timezone(df, year):
...
+        how_merge (str): used for `how=` argument of pd.merge(): e.g. "left", "right",
+        "inner", "outer"
```
rouille, line 355: same comment about indentation

src/oge/helpers.py
```diff
@@ -331,3 +332,49 @@ def add_plant_local_timezone(df, year):
...
+        validate_merge (str): used for `validate=` argument of pd.merge(): eg. "1:1",
+        "1:m","m:1","m:m". Depends on the merge key specified in plant_part_to_map and
+        the structure of `df`
```
rouille, line 358: same comment about indentation

src/oge/load_data.py
```diff
@@ -490,19 +490,27 @@ def update_epa_to_eia_map(cems_df: pd.DataFrame) -> pd.DataFrame:
     return cems_df
 
 
-def add_report_date(df: pd.DataFrame) -> pd.DataFrame:
+def add_report_date(
+    df: pd.DataFrame, plant_timezone: pd.DataFrame | None = None
+) -> pd.DataFrame:
     """Add a report date column to the CEMS data frame based on the plant's local
     timezone
```
rouille, line 497: I guess it applies to all data frame with the required columns. Not only CEMS. Or we should rename the function to `add_report_date_to_cems_df`

src/oge/emissions.py
```diff
@@ -286,48 +296,89 @@ def adjust_fuel_and_emissions_for_CHP(df):
...
     # if the allocation factor < 0, set to zero
-    df.loc[df["electric_allocation_factor"] < 0, "electric_allocation_factor"] = 0
+    df_subplant.loc[
+        df_subplant["electric_allocation_factor"] < 0, "electric_allocation_factor"
+    ] = 0
     # if the allocation factor > 1, set to one
-    df.loc[df["electric_allocation_factor"] > 1, "electric_allocation_factor"] = 1
+    df_subplant.loc[
+        df_subplant["electric_allocation_factor"] > 1, "electric_allocation_factor"
+    ] = 1
```
rouille, line 370: you can use the [`clip`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.clip.html) method here to assign values outside boundary to boundary values.

rouille (approved):
Thanks! I left some docs/style related comments.

src/oge/emissions.py
```diff
@@ -223,8 +223,18 @@ def identify_geothermal_generator_geotype(year):
     return geothermal_geotype
 
 
-def adjust_fuel_and_emissions_for_CHP(df):
-    """Allocates total emissions for electricity generation."""
+def adjust_fuel_and_emissions_for_CHP(df: pd.DataFrame) -> pd.DataFrame:
```
rouille, line 226: Name of function should be in lowercase: CHP -> chp

docs/docs/Methodology/Emissions Calculations/Adjusting Emissions for CHP.md
```diff
@@ -7,6 +7,10 @@ Combined Heat and Power (CHP) plants produce both electricity and useful thermal
 
 The formula and assumptions used for calculating the electric allocation factor are copied directly from the [eGRID2020 methodology](https://www.epa.gov/system/files/documents/2022-01/egrid2020_technical_guide.pdf) and is summarized below (track progress on updates to this methodology [here](https://github.com/singularity-energy/open-grid-emissions/issues/23)).
 
+Although both total `fuel_consumed_mmbtu` and `fuel_consumed_mmbtu_for_electricity` are reported in EIA-923, the eGRID methodology does not directly use these values to calculate an electric allocation factor. Although this is not explicitly explained in the eGRID methodology, this is ostensibly because they are trying to allocate fuel based on the the proportion of energy outputs, not energy inputs. This means that the `fuel_consumed_mmbtu_for_electricity` values calculated using this method will not match the reported `fuel_consumed_mmbtu_for_electricity` values from EIA-923.
```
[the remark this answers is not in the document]
grgmiller: fixed

src/oge/emissions.py
grgmiller: Thanks, I'll fix this throughout

src/oge/load_data.py
grgmiller: Updated

src/oge/emissions.py
grgmiller: I didn't know about clip! Just implemented

src/oge/emissions.py
grgmiller: Fixed

PR #364, opened 2024-05-21
Add geographical information to the plant static attributes data frame

rouille:
### Purpose
Add the following fields to the plant static attributes: "latitude", "longitude", "county", "city", "zip_code", "street_address" and "capacity_mw". Closes CAR-4208.

Add missing typehints and add/format docstrings in modules that were edited.

### What the code is doing
* Create a function to load the geographical information of the plants from the raw EIA-860 data
* Load the `plants_entity_eia` PUDL table to retrieve the geographical information of the plants
* Supplement the PUDL table with data from the raw EIA-860 when entries are missing
* Output the plant static attributes file to disk for years < 2019

### Testing
Successfully ran the 2010 pipeline.

### Where to look
The `oge.load_data` module where the new function load the geographical information from the raw EIA-860 and the `oge.helpers` module where fields to the plant static attributes data frame are added.

### Usage Example/Visuals
Screen shot of the exported 2010 plant static attributes table
Screenshot 2024-05-21 at 1 25 23 AM

### Review estimate
15min

### Future work
CAR-4205

### Checklist

### DIFF
--- a/src/oge/column_checks.py
+++ b/src/oge/column_checks.py
@@ -20,6 +20,8 @@ checks.
 
 from oge.logging_util import get_logger
 
+import pandas as pd
+
 logger = get_logger(__name__)
 
 
@@ -190,11 +192,17 @@ COLUMNS = {
         "fuel_category_eia930",
         "ba_code",
         "ba_code_physical",
-        "state",
         "distribution_flag",
         "timezone",
         "data_availability",
         "shaped_plant_id",
+        "latitude",
+        "longitude",
+        "state",
+        "county",
+        "city",
+        "plant_name_eia",
+        "capacity_mw",
     },
     "plant_metadata": {
         "plant_id_eia",
@@ -364,9 +372,17 @@ COLUMNS = {
 }
 
 
-def check_columns(df, file_name):
-    """
-    Given a file name and a dataframe to export, check that its columns are as expected.
+def check_columns(df: pd.DataFrame, file_name: str):
+    """Given a file name and a data frame to export, check that its columns are as
+    expected.
+
+    Args:
+        df (pd.DataFrame): table to check.
+        file_name (str): key name of file in `COLUMNS`.
+
+    Raises:
+        ValueError: if `file_name` cannot be found in list of files.
+        ValueError: if columns are missing in `df`.
     """
 
     cols = set(list(df.columns))
@@ -491,13 +507,26 @@ def get_dtypes():
         "subplant_primary_fuel_from_net_generation_mwh": "str",
         "timezone": "str",
         "wet_dry_bottom": "str",
+        "latitude": "float64",
+        "longitude": "float64",
+        "county": "str",
+        "city": "str",
+        "plant_name_eia": "str",
     }
 
     return dtypes_to_use
 
 
-def apply_dtypes(df):
-    """Applies specified dtypes to a dataframe and identifies if a dtype is not specified for a column."""
+def apply_dtypes(df: pd.DataFrame) -> pd.DataFrame:
+    """Applies specified types to a data frame and identifies if a type is not
+    specified for a column.
+
+    Args:
+        df (pd.DataFrame): table whose columns will be converted.
+
+    Returns:
+        pd.DataFrame: original data frame with type converted columns.
+    """
     dtypes = get_dtypes()
     datetime_columns = ["datetime_utc", "datetime_local", "report_date"]
     cols_missing_dtypes = [
@@ -507,7 +536,8 @@ def apply_dtypes(df):
     ]
     if len(cols_missing_dtypes) > 0:
         logger.warning(
-            "The following columns do not have dtypes assigned in `column_checks.get_dtypes()`"
+            "The following columns do not have dtypes assigned in "
+            "`column_checks.get_dtypes()`"
         )
         logger.warning(cols_missing_dtypes)
     return df.astype({col: dtypes[col] for col in df.columns if col in dtypes})
--- a/src/oge/data_pipeline.py
+++ b/src/oge/data_pipeline.py
@@ -11,6 +11,8 @@ import argparse
 import os
 import shutil
 
+import pandas as pd
+
 # import local modules
 import oge.download_data as download_data
 import oge.data_cleaning as data_cleaning
@@ -41,13 +43,19 @@ def get_args() -> argparse.Namespace:
     parser.add_argument("--year", help="Year for analysis", default=2022, type=int)
     parser.add_argument(
         "--shape_individual_plants",
-        help="Assign an hourly profile to each individual plant with EIA-only data, instead of aggregating to the fleet level before shaping.",
+        help=(
+            "Assign an hourly profile to each individual plant with EIA-only data, "
+            "instead of aggregating to the fleet level before shaping."
+        ),
         default=True,
         action=argparse.BooleanOptionalAction,
     )
     parser.add_argument(
         "--small",
-        help="Run on subset of data for quicker testing, outputs to outputs/small and results to results/small.",
+        help=(
+            "Run on subset of data for quicker testing, outputs to outputs/small and "
+            "results to results/small."
+        ),
         default=False,
         action=argparse.BooleanOptionalAction,
     )
@@ -93,11 +101,14 @@ def main(args):
     os.makedirs(outputs_folder(f"{path_prefix}"), exist_ok=True)
     os.makedirs(outputs_folder(f"{path_prefix}/eia930"), exist_ok=True)
     if not args.skip_outputs:
-        # If we are outputing, wipe results dir so we can be confident there are no old result files (eg because of a file name change)
+        # If we are outputing, wipe results dir so we can be confident there are no old
+        # result files (eg because of a file name change)
         if os.path.exists(results_folder(f"{path_prefix}")):
             shutil.rmtree(results_folder(f"{path_prefix}"))
         os.makedirs(results_folder(f"{path_prefix}"), exist_ok=False)
-    else:  # still make sure results dir exists, but exist is ok and we won't be writing to it
+    else:
+        # still make sure results dir exists, but exist is ok and we won't be writing
+        # to it
         os.makedirs(results_folder(f"{path_prefix}"), exist_ok=True)
     os.makedirs(
         results_folder(f"{path_prefix}data_quality_metrics"),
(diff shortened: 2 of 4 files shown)


src/oge/data_pipeline.py
```diff
@@ -678,6 +692,15 @@ def main(args):
             ba_fuel_data, year, path_prefix, args.skip_outputs, include_hourly=False
         )
 
+        # export plant static attributes to csv
+        output_data.output_intermediate_data(
+            plant_attributes.assign(shaped_plant_id=np.NaN),
```
grgmiller, line 697: This should be `pd.NA` since shaped_plant_id should have an Int dtype

src/oge/helpers.py
```diff
@@ -174,17 +202,11 @@ def create_plant_ba_table(year):
...
     )
 
-    # convert the dtype of the balancing authority code column from string to object
-    # this will allow for missing values to be filled
-    plant_ba["balancing_authority_code_eia"] = plant_ba[
-        "balancing_authority_code_eia"
-    ].astype(object)
-    plant_ba["balancing_authority_code_eia"] = plant_ba[
-        "balancing_authority_code_eia"
-    ].fillna(value=np.NaN)
```
grgmiller, line 184: I'm assuming you removed this to have more consistent dtypes, but did you confirm that the original behavior is as expected? ie that missing BAs are getting filled properly?

src/oge/helpers.py
```diff
@@ -174,17 +202,11 @@ def create_plant_ba_table(year):
...
-
     # load the ba name reference
-    ba_name_to_ba_code = pd.read_csv(reference_table_folder("ba_reference.csv"))
+    ba_name_to_ba_code = pd.read_csv(
+        reference_table_folder("ba_reference.csv"),
+        dtype={"ba_name": "string", "ba_code": "string"},
```
grgmiller, line 208: should we just use `column_checks.get_dtypes()` here?

src/oge/helpers.py
```diff
@@ -194,7 +216,8 @@ def create_plant_ba_table(year):
 
     # specify a ba code for certain utilities
     utility_as_ba_code = pd.read_csv(
-        reference_table_folder("utility_name_ba_code_map.csv")
+        reference_table_folder("utility_name_ba_code_map.csv"),
+        dtype={"name": "string", "ba_code": "string"},
```
grgmiller, line 220: should we just use column_checks.get_dtypes() here?

src/oge/helpers.py
```diff
@@ -229,7 +253,8 @@ def create_plant_ba_table(year):
             retired_bas["retirement_date"].dt.year < year, "ba_code"
         ].unique()
     )
-    # if there are any plants that have been assigned to a retired BA, set its BA code as missing
+    # if there are any plants that have been assigned to a retired BA, set its BA code
+    # as missing
     plant_ba.loc[plant_ba["ba_code"].isin(retired_bas), "ba_code"] = np.NaN
```
grgmiller, line 258: To be consistent with the deleted code above, we should use `pd.NA` here

src/oge/helpers.py
```diff
@@ -325,11 +389,49 @@ def assign_fuel_category_to_ESC(
...
+            'state', 'county', 'city', 'zip_code' and 'street_address' columns
+    """
+    geographical_info = [
+        "latitude",
+        "longitude",
+        "state",
+        "county",
+        "city",
+        "zip_code",
+        "street_address",
```
grgmiller, line 409: I would propose removing street address and zip code, but adding the `plant_name_eia` column.
I don't think we need street address since we have lat/long info

src/oge/helpers.py
```diff
@@ -325,11 +389,49 @@ def assign_fuel_category_to_ESC(
...
+    ]
+    plants_entity = load_data.load_pudl_table(
+        "plants_entity_eia", columns=["plant_id_eia", "timezone"] + geographical_info
+    ).astype({"zip_code": "Int64"})
+    plants_entity_from_eia860 = load_data.load_raw_eia860_plant_geographical_info(
+        latest_validated_year
```
grgmiller, line 415: I get why you use latest_validated_year here, but do we download this file for the latest validated year no matter what year we're running? Just want to make sure we don't run into a filenotfounderror.

src/oge/helpers.py
```diff
@@ -257,22 +282,55 @@ def create_plant_ba_table(year):
...
+        "generators_eia860",
+        columns=["plant_id_eia", "generator_id", "report_date", "capacity_mw"],
+    )
+    plants_capacity = (
+        generators_capacity.query(
+            "report_date == @generators_capacity['report_date'].max()"
+        )
+        .groupby(["plant_id_eia"])["capacity_mw"]
+        .sum()
+        .round(2)
```
grgmiller, line 309: Why is rounding necessary? Do we ever have data with more than 2 decimals?
We may want to keep the original level of detail

src/oge/helpers.py
```diff
@@ -257,22 +282,55 @@ def create_plant_ba_table(year):
...
+        pd.DataFrame: original data frame with additional 'capacity_mw' column.
     """
-    Identifies which plant_id_eia are "distribution grid connected" based on a voltage threshold.
+    generators_capacity = load_data.load_pudl_table(
+        "generators_eia860",
+        columns=["plant_id_eia", "generator_id", "report_date", "capacity_mw"],
+    )
+    plants_capacity = (
+        generators_capacity.query(
+            "report_date == @generators_capacity['report_date'].max()"
```
grgmiller, line 305: Why do we use this query format here rather than just: `generators_capacity[generators_capacity["report_date"] == generators_capacity["report_date"].max()]]`

grgmiller (changes requested):
See comments and requested changes. 

I'm proposing that we add the plant name to this file, and drop the street address/zip code unless you think we will use those.

src/oge/helpers.py
rouille: `ba_name` is not in `get_types()`. Should I add it?

src/oge/helpers.py
rouille: Same issue, `name` is not in `get_dtypes()`. This is more tricky since a key named name is ambiguous. I would change the name of the header of the first column in the **utility_name_ba_code_map.csv** file to `utility_name`.

src/oge/helpers.py
rouille: Sounds good

src/oge/helpers.py
rouille: This is not an issue when running the pipeline since the EIA-860 files are downloaded for the latest validated year in Step 1.

Now if you believe we could use this function outside of the pipeline I can introduce a logic where we check if the file exists and if not is downloaded before loading it.

src/oge/helpers.py
rouille: Most values have only digit after the decimal point. The ones that don't have one digit after the decimal point followed by 9s, e.g., 3.5999999999999999

src/oge/helpers.py
rouille: I did compare for 2022 the plant static attributes files before implementing the code in this PR and after:
```
>>> old[["plant_id_eia", "ba_code"]].equals(new[["plant_id_eia", "ba_code"]])
True
>>> old[["plant_id_eia", "ba_code", "state"]].compare(new[["plant_id_eia", "ba_code", "state"]])
      state      
       self other
797     NaN    WI
1924    NaN    VT
4319    NaN    MN
11630   NaN    MD
```
We can see that `ba_code` did not change between the two versions and that the 4 rows were `state` was missing are now filled out.

grgmiller: sounds good

src/oge/helpers.py
grgmiller: Thanks, this is fine as is then!

src/oge/helpers.py
grgmiller: ok thanks, this makes sense then

src/oge/helpers.py
grgmiller: Maybe it doesn't matter but it's just not a pattern that we use in the rest of the repo. Either way is fine. It looks like there shouldn't be any major performance difference between the two based on some quick googling

src/oge/helpers.py
grgmiller: No, this is fine. I generally like using our built in function since we have it, but I don't think that there's really a benefit to using it here.

src/oge/helpers.py
grgmiller: Probably no need here unless you really want to

grgmiller (approved):
I'm going to go ahead and approve assuming that you will be making the changes discussed in the previous review just so I don't hold up your progress.

rouille:
> I'm going to go ahead and approve assuming that you will be making the changes discussed in the previous review just so I don't hold up your progress. 

Thanks. I launched the pipeline overnight to make sure the changes did not break it. Il will check the result when I am at my desk and commit my changes if it looks good.

src/oge/helpers.py
rouille: Done

Screenshot 2024-05-22 at 9 46 54 AM

src/oge/column_checks.py
```diff
@@ -364,9 +373,17 @@
...
 
-def check_columns(df, file_name):
-    """
-    Given a file name and a dataframe to export, check that its columns are as expected.
+def check_columns(df: pd.DataFrame, file_name: str):
+    """Given a file name and a data frame to export, check that its columns are as
+    expected.
+
+    Args:
+        df (pd.DataFrame): table to export.
```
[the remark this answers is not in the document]
rouille: Done

src/oge/column_checks.py
```diff
@@ -364,9 +373,17 @@
...
+        file_name (str): template to check against.
```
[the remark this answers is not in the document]
rouille: Done

src/oge/data_pipeline.py
rouille: Done

src/oge/helpers.py
rouille: It can be done in a quick PR later. We have bigger fishes to fry right now.

src/oge/helpers.py
rouille: Done

src/oge/helpers.py
rouille: Changed.

PR #367, opened 2024-05-22
Add operating and retirement dates to plant static attributes

rouille:
### Purpose
Add operating and retirement dates to plant static attributes.

A bug is fixed when calculating the nameplate capacity at the plant level

This PR also fixes an issue for years < 2013 where missing BA codes were being assigned, resulting in inaccurate BA-level results.

### What the code is doing
Create a new function that adds the operating and retirement dates of a plant to the plant static attributes data frame. The operating date of a plant is taken as the earliest date among all generators' operating date over all report dates. Likewise, the retirement date of a plant is taken as the latest date among all generators' retirement date over all report dates.

### Testing
Successfully ran the 2013 pipeline. 

### Where to look
* the new `add_plant_operating_and_retirement_dates` in the `oge.helpers` module.
* the `oge.column_checks` module where the new fields were added. Note that I added some missing datetime to the list of columns defined in the `apply_dtypes` function. These columns won't be converted.

### Usage Example/Visuals
Screenshot 2024-05-22 at 7 28 14 PM

### Review estimate
10min

### Future work
N/A

### Checklist

### DIFF
--- a/src/oge/column_checks.py
+++ b/src/oge/column_checks.py
@@ -203,6 +203,8 @@ COLUMNS = {
         "city",
         "plant_name_eia",
         "capacity_mw",
+        "plant_operating_date",
+        "plant_retirement_date",
     },
     "plant_metadata": {
         "plant_id_eia",
@@ -422,6 +424,7 @@ def get_dtypes():
         "ch4_mass_lb_adjusted": "float64",
         "ch4_mass_lb_for_electricity": "float64",
         "ch4_mass_lb_for_electricity_adjusted": "float64",
+        "city": "str",
         "co2_mass_lb": "float64",
         "co2_mass_lb_adjusted": "float64",
         "co2_mass_lb_for_electricity": "float64",
@@ -431,6 +434,7 @@ def get_dtypes():
         "co2e_mass_lb_adjusted": "float64",
         "co2e_mass_lb_for_electricity": "float64",
         "co2e_mass_lb_for_electricity_adjusted": "float64",
+        "county": "str",
         "data_availability": "category",
         "distribution_flag": "bool",
         "eia930_profile": "float64",
@@ -455,6 +459,8 @@ def get_dtypes():
         "hourly_data_source": "category",
         "hours_in_service": "float64",
         "imputed_profile": "float64",
+        "latitude": "float64",
+        "longitude": "float64",
         "mercury_control_id_eia": "str",
         "mercury_emission_rate_lb_per_trillion_btu": "float64",
         "mercury_removal_efficiency": "float64",
@@ -478,6 +484,7 @@ def get_dtypes():
         "particulate_removal_efficiency_at_full_load": "float64",
         "plant_id_eia": "Int32",
         "plant_id_epa": "Int32",
+        "plant_name_eia": "str",
         "plant_primary_fuel": "str",
         "plant_primary_fuel_from_capacity_mw": "str",
         "plant_primary_fuel_from_fuel_consumed_for_electricity_mmbtu": "str",
@@ -507,11 +514,6 @@ def get_dtypes():
         "subplant_primary_fuel_from_net_generation_mwh": "str",
         "timezone": "str",
         "wet_dry_bottom": "str",
-        "latitude": "float64",
-        "longitude": "float64",
-        "county": "str",
-        "city": "str",
-        "plant_name_eia": "str",
     }
 
     return dtypes_to_use
@@ -528,7 +530,16 @@ def apply_dtypes(df: pd.DataFrame) -> pd.DataFrame:
         pd.DataFrame: original data frame with type converted columns.
     """
     dtypes = get_dtypes()
-    datetime_columns = ["datetime_utc", "datetime_local", "report_date"]
+    datetime_columns = [
+        "datetime_utc",
+        "datetime_local",
+        "report_date",
+        "generator_operating_date",
+        "generator_retirement_date",
+        "current_planned_generator_operating_date",
+        "plant_operating_date",
+        "plant_retirement_date",
+    ]
     cols_missing_dtypes = [
         col
         for col in df.columns
--- a/src/oge/helpers.py
+++ b/src/oge/helpers.py
@@ -2,7 +2,7 @@ import numpy as np
 import pandas as pd
 
 from oge.column_checks import get_dtypes, apply_dtypes
-from oge.constants import latest_validated_year
+from oge.constants import earliest_data_year, latest_validated_year
 from oge.filepaths import reference_table_folder, outputs_folder
 import oge.load_data as load_data
 from oge.logging_util import get_logger
@@ -122,6 +122,9 @@ def create_plant_attributes_table(
 
     # add nameplate capacity
     plant_attributes = add_plant_nameplate_capacity(plant_attributes)
+
+    # add operating and retirement dates
+    plant_attributes = add_plant_operating_and_retirement_dates(plant_attributes)
 
     plant_attributes = apply_dtypes(plant_attributes)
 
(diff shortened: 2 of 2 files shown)


src/oge/column_checks.py
```diff
@@ -528,7 +530,16 @@ def apply_dtypes(df: pd.DataFrame) -> pd.DataFrame:
...
     dtypes = get_dtypes()
-    datetime_columns = ["datetime_utc", "datetime_local", "report_date"]
+    datetime_columns = [
+        "datetime_utc",
+        "datetime_local",
+        "report_date",
+        "generator_operating_date",
+        "generator_retirement_date",
+        "current_planned_generator_operating_date",
+        "operating_date",
```
grgmiller, line 540: to follow the pudl naming conventions, and for clarity, let's call these "plant_operating_date" and "plant_retirement_date"

src/oge/helpers.py
```diff
@@ -287,6 +290,52 @@ def create_plant_ba_table(year: int) -> pd.DataFrame:
...
+            "report_date",
+            "generator_operating_date",
+            "generator_retirement_date",
+        ],
+    )
+    plants_dates = (
+        generators_dates.groupby("plant_id_eia")[
+            ["generator_operating_date", "generator_retirement_date"]
+        ]
+        .agg({"generator_operating_date": "min", "generator_retirement_date": "max"})
```
grgmiller, line 325: While this will work for the operating date, this will not work for the retirement date. For example, what if a plant has 10 generators and only one retires? This would currently say the entire plant is retired.

For the retirement date, one way to do this would be to check for plants where there are no NA retirement dates across all generators, and then take the max of that.

Looking at the sample outputs you posted, it currently shows that plant 3 "Barry" retired in 2015, but this plant is still operational

src/oge/helpers.py
```diff
@@ -287,6 +290,52 @@ def create_plant_ba_table(year: int) -> pd.DataFrame:
...
+    generator.
+
+    Args:
+        df (pd.DataFrame): table with a 'plant_id_eia' column.
+
+    Returns:
+        pd.DataFrame: original data frame with additional 'operating_date' and
+            'retirement_date' column.
+    """
+    generators_dates = load_data.load_pudl_table(
```
grgmiller, line 311: "generator_dates"

src/oge/helpers.py
```diff
@@ -287,6 +290,52 @@ def create_plant_ba_table(year: int) -> pd.DataFrame:
...
+        "denorm_generators_eia",
```
grgmiller, line 312: this table will contain values for each year reported, so before we run our min and max operations, we need to drop duplicates. Before we drop duplicates though, we may need to do a groupby([plant_id, generator_id]).ffill() and .bfill() to make sure that we have complete values for all years

src/oge/helpers.py
grgmiller: One easy way to do this would be to load the "operational_status' column and just identify where all generators are retired as of the latest_validated_year

src/oge/helpers.py
grgmiller: We may also want to filter to only include data up to the latest_validated_year

grgmiller: This is something I was working on in GRETA, but a similar pattern may work here:
```
min_operating = oge.load_data.load_pudl_table(
    "generators_eia860",
    year=earliest_data_year,
    end_year=latest_validated_year,
    columns=[
        "report_date",
        "plant_id_eia",
        "generator_id",
        "minimum_load_mw",
        "capacity_mw",
        "summer_capacity_mw",
        "winter_capacity_mw",
    ],
).sort_values(by=["plant_id_eia","generator_id","report_date"], ascending=True)

# fill missing capacity values
capacity_columns = ["minimum_load_mw", "capacity_mw", "summer_capacity_mw", "winter_capacity_mw"]

for col in capacity_columns:
    min_operating[col] = min_operating.groupby(["plant_id_eia","generator_id"])[col].bfill()
    min_operating[col] = min_operating.groupby(["plant_id_eia","generator_id"])[col].ffill()

# keep only the most recent year of data
min_operating = min_operating.drop_duplicates(subset=["plant_id_eia","generator_id"], keep="last")
```

grgmiller (changes requested):
See requested changes. 

Before we merge this, it may be helpful to do a couple quick validations:
- Are there any missing operating dates? If so, we will just need to understand how to deal with those in MISO
- Load up some EIA-923 data for this year and just quickly check if there are any plants that we marked as retired prior to 2022 that reported 923 data in 2022... sometimes plants do continue to report, so this may be okay, but we should at least manually double check that our algorithm didn't mistakenly mark a plant as retired if at least one generator is still going

src/oge/helpers.py
rouille: Indeed

src/oge/column_checks.py
rouille: Done

src/oge/helpers.py
rouille: Done

src/oge/helpers.py
rouille: Changed

src/oge/helpers.py
rouille: Thanks. Implemented.

rouille:
Screen shot with new implementation
Screenshot 2024-05-23 at 9 30 11 AM

grgmiller (approved):
Looks good, thanks. 
One small request would be to change the order of the columns so that we are grouping data together and make it easier to read. My suggestion for column order would be:
```
      "plant_id_eia", #identification columns
      "plant_name_eia",
       "capacity_mw", # what type of plant is this
      "plant_primary_fuel",
      "fuel_category",
      "fuel_category_eia930",
      "state", # where is it located
        "county",
        "city",
      "ba_code",
      "ba_code_physical",
     "latitude",
      "longitude",
      "distribution_flag", #other random metadata
      "timezone",
      "data_availability",
      "shaped_plant_id",
      
        
        
```

rouille:
The calculation of the nameplate capacity was bugged and is fixed in the katest commit

rouille:
> Looks good, thanks. One small request would be to change the order of the columns so that we are grouping data together and make it easier to read. My suggestion for column order would be:
> 
> ```
>       "plant_id_eia", #identification columns
>       "plant_name_eia",
>        "capacity_mw", # what type of plant is this
>       "plant_primary_fuel",
>       "fuel_category",
>       "fuel_category_eia930",
>       "state", # where is it located
>         "county",
>         "city",
>       "ba_code",
>       "ba_code_physical",
>      "latitude",
>       "longitude",
>       "plant_operating_date", #operational status columns
>       "plant_retirement_date",
>       "distribution_flag", #other random metadata
>       "timezone",
>       "data_availability",
>       "shaped_plant_id",
>       
>         
>         
> ```

Done

src/oge/helpers.py
```diff
@@ -296,21 +393,33 @@ def add_plant_nameplate_capacity(df: pd.DataFrame) -> pd.DataFrame:
...
+
+    generator_capacity["capacity_mw"] = generator_capacity.groupby(
+        ["plant_id_eia", "generator_id"]
+    )["capacity_mw"].bfill()
+    generator_capacity["capacity_mw"] = generator_capacity.groupby(
+        ["plant_id_eia", "generator_id"]
+    )["capacity_mw"].ffill()
+
+    # keep only the most recent year of data
+    generator_capacity = generator_capacity.drop_duplicates(
```
grgmiller, line 411: In the case of nameplate capacity, I think that we only want to keep the specific data year, not the latest validated year. Nameplate capacity can chance over time if the generator is repowered, so this value might be annually varying.

grgmiller: It's still good that we load all years and do the fill in case there is missing capacity data in a specific year.

grgmiller (commented):
See comment about the nameplate capacity fix.

grgmiller (approved):
Capacity changes look good

PR #369, opened 2024-06-20
Update to use most recent version of pudl

grgmiller:
### Purpose
This PR updates OGE to use pudl v2024.5.0 release data. This primarily involves updating table names.

Note, that we are still using an older version of the pudl dependency, but since all of the column names appear to be the same, this is not causing any errors for now. However, in a future PR (see future work section), we should strive to just completely remove the pudl code dependency from OGE and rely entirely on the output data. I did start to remove some dependencies where it was easy, and added some notes for the future:
- In `emissions.py`: we import a specific function `pudl.analysis.allocate_gen_fuel.distribute_annually_reported_data_to_months_if_annual()`, which takes data with a single annual report date, and expands it to each month in that year. However, after examining what this function is actually doing, the new pudl table already has this functionality included, so we do not need to run it again.
- - In `load_data.py`: we import `from pudl.metadata.fields import apply_pudl_dtypes`. We use this to apply the pudl-defined dtypes to each column of data loaded from the pudl sqlite database. To remove this, I've updated `column_checks` to include a dictionary of pudl dtypes that we apply

Closes CAR-4361.

### Testing
Ran pipeline successfully for 2018.
Ran pipeline for 2019-2022, including download of new data source.

### Where to look
N/A

### Review estimate
10 min

### Future work
In a future PR, we should see if we can completely remove the PUDL dependency from OGE. We may be able to do this using the new tables that are available. The current dependencies are:

- In `subplant_identification.py`: 
     - We `from pudl.etl.glue_assets import make_subplant_ids`: we use this to run our own pipeline for creating subplant_ids. Replacing this would likely require us to integrate our subplant pipeline into pudl. 
     - We `import pudl.analysis.epacamd_eia as epacamd_eia`, we use this to `filter_crosswalk()`, again as part of creating our own subplant_ids
- in `data_cleaning.py`: we  `import pudl.analysis.allocate_gen_fuel as allocate_gen_fuel` to run our own allocation. We could easily replace this but we have some code changes in our version of pudl that we'd need to merge. See changes in https://github.com/catalyst-cooperative/pudl/compare/main...singularity-energy:pudl:oge_release primarily from https://github.com/singularity-energy/pudl/pull/1 and https://github.com/singularity-energy/pudl/pull/3

We should also switch to using the pudl emissions control data instead of `load_data.load_emissions_controls_eia923()` once https://github.com/catalyst-cooperative/pudl/issues/3689 is completed.

### Checklist


src/oge/data_cleaning.py
```diff
@@ -72,13 +72,11 @@ def clean_eia923(
...
-        "denorm_generation_fuel_combined_monthly_eia923", year
-    )
-    bf = load_data.load_pudl_table("denorm_boiler_fuel_monthly_eia923", year)
-    gen = load_data.load_pudl_table("denorm_generation_monthly_eia923", year)
-    gens = load_data.load_pudl_table("denorm_generators_eia", year)
-    bga = load_data.load_pudl_table("boiler_generator_assn_eia860", year)
+    gf = load_data.load_pudl_table("out_eia923__monthly_generation_fuel_combined", year)
+    bf = load_data.load_pudl_table("out_eia923__monthly_boiler_fuel", year)
+    gen = load_data.load_pudl_table("out_eia923__monthly_generation", year)
+    gens = load_data.load_pudl_table("_out_eia__yearly_generators", year)
```
rouille, line 78: I guess you checked that it is correct but I ask because it looks weird, should `_out` be `out` ?

src/oge/helpers.py
```diff
@@ -366,7 +366,7 @@ def add_plant_operating_and_retirement_dates(df: pd.DataFrame) -> pd.DataFrame:
             'plant_retirement_date' column.
     """
     generator_dates = load_data.load_pudl_table(
-        "denorm_generators_eia",
+        "out_eia__yearly_generators",
```
rouille, line 369: In `oge.data_cleaning`, it is `_out_eia__yearly_generators`. See this [comment](https://github.com/singularity-energy/open-grid-emissions/pull/369/files#r1648210955).

src/oge/data_cleaning.py
grgmiller: I am just using the specific table used by pudl in their version of the pipeline here: https://github.com/catalyst-cooperative/pudl/blob/b096fbc63bd5cb5f28c7d6496c618b8a36aa46bd/src/pudl/analysis/allocate_gen_fuel.py#L223. 

I do not completely understand the difference between `_out_eia__yearly_generators` and `out_eia__yearly_generators`

src/oge/helpers.py
grgmiller: See above response. I use the _ version in the allocation pipeline because that is what is used by pudl there, but here I want to use the more general version of this table.

src/oge/data_cleaning.py
grgmiller: See: 
image

rouille: It looks like these tables (the intermediate starting with `_`) will be removed from the database soon. Do we know what to do when this happens?

grgmiller: I can ask

grgmiller: Looking at the differences between these two tables:
image
It looks like they have the same number of rows, but the new table has 8 more columns than in the legacy table. I think based on this, it would be safe to switch to the new table.

grgmiller:
Re-requesting your review since I added a lot of new changes related to removing pudl dependencies @rouille

src/oge/column_checks.py
```diff
@@ -530,25 +778,50 @@ def apply_dtypes(df: pd.DataFrame) -> pd.DataFrame:
...
+
+    # apply the dtype if the column is in the dtype dict
+    df = df.astype({col: dtypes[col] for col in df.columns if (col in dtypes)})
+    # apply datetime columns
+    for col in datetime_cols:
+        if col in df.columns:
+            try:
+                df[col] = df[col].astype("datetime64[s]")
+            except TypeError:
+                df[col] = df[col].dt.tz_localize(None).astype("datetime64[s]")
```
rouille, line 825: When does this happen, i.e., that a column in `datetime_cols` cannot be converted to `datetime64[s]`? Just being curious

src/oge/emissions.py
```diff
@@ -1000,11 +995,6 @@ def return_monthly_plant_fuel_heat_content(year):
             "report_date",
             "fuel_mmbtu_per_unit",
         ],
-    ).pipe(
-        distribute_annually_reported_data_to_months_if_annual,
-        key_columns=["plant_id_eia", "energy_source_code", "prime_mover_code"],
-        data_column_name="fuel_mmbtu_per_unit",
-        freq="MS",
```
rouille, line 1007: Why is that not needed anymore?

rouille: Ok, It is explained in the PR description.

src/oge/column_checks.py
grgmiller: This was an issue when there was a dataframe with a datetime_utc column that had already been made tzaware (eg the dtype is `datetime64[s. UTC]`

PR #370, opened 2024-06-21
Fix issues with anomalous gross to net conversions

grgmiller:
### Purpose
This PR addresses an issue where we were noticing anomalous net generation data for certain generators in 2005. Specifically, the generation was several orders of magnitude greater than the reported gross generation or net generation. This was a result of high backstop values and a lack of validation of these values. 

To fix this specific issue, I updated how the `annual_fuel_ratio` (now renamed `annual_fleet_ratio`) is calculated. Previously, we were just taking the mean GTN value for all plants that had a specific energy_source_code as the primary fuel. However, this approach was subject to outliers. We now filter anomalous subplants out of the data first, and then calculate a weighted average value across all subplants. Instead of using the specific energy_source_code, we now calculate fleet ratios for fuel_category, prime_mover_code combinations. Including the prime_mover_code is more consistent with the EIA backstop/default GTN values, which are organized by prime mover. 

While digging into this, I also took the opportunity to address other issues with the GTN methodology. 

- This PR also removes the "shift_factor" methods from the method hierarchy. These approaches were only used by a small portion of the data, and their methodology is less robust than the ratio method. 
- I also addressed an issue where prime_mover_codes were not being added to all generators, meaning that default factors could not be added for some generators. 
- Another issue was that some data was being summed for subplant-months with only EIA data or only CEMS data, but not both, which was resulting in abnormally high or low factors. I added in extra filtering to ensure that we are only including rows that contain both CEMS and EIA data when calculating GTN ratios.
- I also updated the documentation to better describe the current GTN methodology.

**Note**: This PR also fixes an issue not addressed in https://github.com/singularity-energy/open-grid-emissions/pull/369 about setting the datetime dtype for columns that were already tzaware datetimes.

### Testing
Ran the GTN pipeline for 2022, and ran the entire pipeline for 2005 data.

**Comparing the old 2005-2006 data for MISO:**
Previous carbon intensity: 1,530 lb/MWh (2005), 1,650 lb/MWh (2006)
After GTN fix: 1,717 lb/MWh (2005), 1,691 lb/MWh

### Where to look

### Usage Example/Visuals

### Review estimate

### Future work

### Checklist


src/oge/gross_to_net_generation.py
```diff
@@ -193,19 +183,19 @@ def calculate_gross_to_net_conversion_factors(
...
-            validate="m:1",
-        )
         .merge(
             net_gen_data,
             how="outer",
             on=["plant_id_eia", "subplant_id", "report_date"],
             indicator="data_source",
             validate="1:1",
         )
+        .merge(
```
grgmiller, line 193: Moved this to after merging in the EIA data so that prime mover codes would be added for all generators.

docs/docs/Methodology/Converting Gross to Net Generation.md
```diff
@@ -54,12 +54,11 @@ The conversion factors are applied in the following hierarchical order:
 
 1. Subplant gross-to-net ratio
 2. Plant gross-to-net ratio
-3. Subplant shift factor
-4. Plant shift factor
-5. Fuel-specific ratio
-6. Set net generation equal to net generation
+3. Fleet-specific ratio
+4. Default GTN ratios from EIA 
+5. Set net generation equal to net generation
```
rouille, line 59: Set net net generation equal to gross generation?

docs/docs/Methodology/Converting Gross to Net Generation.md
```diff
@@ -44,22 +44,21 @@ Following the approach primarily used in previous academic literature, the OGE d
...
+Ratios are calculated both at the subplant and plant level and both at the monthly and annual resolutions. Monthly-resolution conversion factors are not currently used in the data pipeline because we are not always confident that the monthly values reported in EIA-923 are accurate. For example, in some cases, generators report annual totals and EIA distributes the data to each month. The use of annual-resolution conversion factors means that the annual total calculated net generation will match the annual total reported net generation, but monthly total calculated net generation will not always match monthly total reported net generation.
```
rouille, line 47: First sentence. Should `resolutions` has no `s` or should `level` has one? I would say both should have no `s`

rouille (approved):
LGTM

docs/docs/Methodology/Converting Gross to Net Generation.md
grgmiller: Updated, we actually no longer use this method

docs/docs/Methodology/Converting Gross to Net Generation.md
grgmiller: Fixed.

src/oge/column_checks.py
```diff
@@ -823,5 +819,7 @@ def apply_dtypes(df: pd.DataFrame) -> pd.DataFrame:
                 df[col] = df[col].astype("datetime64[s]")
             except TypeError:
                 df[col] = df[col].dt.tz_localize(None).astype("datetime64[s]")
+            if "_utc" in col:
+                df[col] = df[col].dt.tz_localize("UTC")
```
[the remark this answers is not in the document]
grgmiller: I think that we want to only apply the UTC tz after the dtype has been set to datetime64, so we'd want to run it after.

PR #368, opened 2024-06-18
Fix and add information to plant static attributes

rouille:
### Purpose
Fix positive longitudes and erroneous time zones of plants, add missing coordinates and location (state, county and city) of plants and save the plant static attributes to the **open_grid_emissions_data/results/plant_data** folder. Closes CAR-4339, CAR-4340, CAR-4341.

### What the code is doing
* Use raw EIA-860 file to fix positive longitude and use the `timezonefinder` package to derive the timezone after fixing the coordinates
* Use `geopy` package to get latitude and longitude of plant from its location (state, county and city) and vice versa

### Testing
Successfully ran the 2005 pipeline

### Where to look
The `oge.helpers` module has most of the changes

### Usage Example/Visuals
Comparing 2005 plant static attributes files before changes and after.

The code snippet shows that erroneous longitude and timezone are fixed
```
>>> psa_old = pd.read_csv("~/Desktop/2005_outputs/plant_static_attributes_2005.csv", usecols=["plant_id_eia", "latitude", "longitude", "state", "county", "city", "timezone"], index_col=0)
>>> psa_new = pd.read_csv("~/open_grid_emissions_data/results/2005/plant_data/plant_static_attributes.csv", usecols=["plant_id_eia", "latitude", "longitude", "state", "county", "city", "timezone"], index_col=0)
>>> psa_old[psa_old["longitude"] > 0]
             state          county            city   latitude   longitude             timezone
plant_id_eia                                                                                  
50060           MD  Prince Georges  Upper Marlboro  38.824863   76.772200          Asia/Urumqi
54537           WA         Whatcom        Ferndale  48.828996  122.686666  America/Los_Angeles
55964           MD  Prince Georges  Upper Marlboro  38.847585   76.776400          Asia/Urumqi
>>> psa_new.loc[psa_old[psa_old["longitude"] > 0].index]
             state          county            city   latitude   longitude             timezone
plant_id_eia                                                                                  
50060           MD  Prince Georges  Upper Marlboro  38.824863  -76.773149     America/New_York
54537           WA         Whatcom        Ferndale  48.828996 -122.685114  America/Los_Angeles
55964           MD  Prince Georges  Upper Marlboro  38.847585  -76.785912     America/New_York
```
The code snippet below shows that most of the missing coordinates and location have been fille out:
```
>>> psa_old.isna().sum()
state         1
county       63
city         12
latitude     46
longitude    30
timezone      0
dtype: int64
>>> psa_new.isna().sum()
state         0
county        4
city         12
[...]

PR #371, opened 2024-06-25
Fix function calculating averages of the fuel types

rouille:
### Purpose
Fix function that calculates the averages of the fuel types. Closes CAR-4205.

### What the code is doing
Add a new row that summarizes all fuel categories as follows:
* Calculate the overall generated emission rates as $\frac{\sum_f E_f}{\sum_f G_f}$, where $E_f$ is the absolute emission of fuel $f$ and $G_f$ is the corresponding net generation
* Take the average of the non-rate columns
* Use `average` as fuel category for this new summarized row. 

### Testing
Ran the 2005 pipeline

### Where to look
Relevant changes are in the the `write_generated_averages` function. Other edits add typehints and docstrings to function in the `oge.output_data` module.

### Usage Example/Visuals
N/A

### Review estimate
5min

### Future work
N/A

### Checklist

### DIFF
--- a/src/oge/output_data.py
+++ b/src/oge/output_data.py
@@ -50,11 +50,12 @@ UNIT_CONVERSIONS = {
 }
 
 
-def prepare_files_for_upload(years):
-    """
-    Zips files in preparation for upload to cloud storage and Zenodo.
-
+def prepare_files_for_upload(years: list):
+    """Zips files in preparation for upload to cloud storage and Zenodo.
     This should only be run when releasing a new minor or major version of the repo.
+
+    Args:
+        years (list): list of four-digit year indicating when the data were taken.
     """
 
     for year in years:
@@ -62,9 +63,13 @@ def prepare_files_for_upload(years):
         zip_data_for_zenodo(year)
 
 
-def zip_results_for_s3(year):
-    """
-    Zips results directories that contain more than a single file for hosting on an Amazon S3 bucket.
+def zip_results_for_s3(year: int):
+    """Zips results directories that contain more than a single file for hosting on an
+    Amazon S3 bucket.
+
+
+    Args:
+        year (int): a four-digit year indicating when the data were taken.
     """
     os.makedirs(data_folder("s3_upload"), exist_ok=True)
     for data_type in ["power_sector_data", "carbon_accounting", "plant_data"]:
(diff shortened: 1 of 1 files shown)


grgmiller:
@rouille could you please post some example screenshots of the new outputs with this fix just to double check that everything is being calculated correctly?

grgmiller (approved):
Looks good but would be nice to see a screenshot of outputs just to confirm

rouille:
> @rouille could you please post some example screenshots of the new outputs with this fix just to double check that everything is being calculated correctly?

Here it is for 2005. There is the `nan` row coming from plant 13213 in Mississippi, see plant attributes below
```
>>> psa = pd.read_csv("plant_static_attributes_2005.csv.zip")
>>> psa[psa["fuel_category"].isna()]
      plant_id_eia       plant_name_eia  capacity_mw plant_primary_fuel fuel_category fuel_category_eia930 state county city ba_code ba_code_physical  latitude  longitude plant_operating_date plant_retirement_date  distribution_flag         timezone data_availability  shaped_plant_id
3389         13213  BTEC New Albany LLC          NaN                NaN           NaN                  NaN    MS    NaN  NaN     NaN              NaN   34.5411   -88.9422                  NaN                   NaN               True  America/Chicago         cems_only              NaN
```
It is fixed in #368 and will be taken care of once we rebase this PR. Nevertheless the calculation is correct. The data frame being to wide, I drop the file instead of a screenshot.
[annual_generation_averages_by_fuel_2005.csv](https://github.com/user-attachments/files/15975426/annual_generation_averages_by_fuel_2005.csv)

grgmiller (changes requested):
So looking at the sample outputs you posted, I think we want to keep the summary row as "total" and not "average". We want to take the sum of all of the numeric columns, and not the average, and it looks like the outputs are still averages of all of the other fuels
image

Could you please change "average" back to "total" and sum the numeric columns instead of averaging them?

I think the others where we reference this data are looking for the "total" row so changing the name will break those.

grgmiller:
The "generated" rate columns are now correct, but we just want to also fix the numerical columns

rouille:
> The "generated" rate columns are now correct, but we just want to also fix the numerical columns



> The "generated" rate columns are now correct, but we just want to also fix the numerical columns

Should we change the file name then, `annual_generation_averages_by_fuel_2005.csv` is confusing in my opinion. I would expect to find average value for all columns including the absolute ones.

grgmiller:
> Should we change the file name then
I think here the averages refers to fuel-average emissions factors, rather than averages across fuels. The total row is a sum (just like each of the fuel rows are a sum of specific plants - we are not calculating the average net generation across all coal plants), and it is the generated_rate columns that are the averages of those sums. 

If we change the file name, we would just need to track down where we are using this and change the file name there as well.

rouille:
> The "generated" rate columns are now correct, but we just want to also fix the numerical columns

Done. See file attached
[annual_generation_averages_by_fuel_2005.csv](https://github.com/user-attachments/files/15976220/annual_generation_averages_by_fuel_2005.csv)

grgmiller (approved):
Looks good, thanks!

Issue #276, opened 2023-02-03
Standardize SO2 and NOx emission factor creation


grgmiller:
The uncontrolled SO2 and NOx emission factor tables that we use are manually created based on two tables included as part of the EIA's Electric Power Annual. However, in copying these factors, applying them to prime movers that are not explicitly listed, and applying rules to determine which default value to use, there is a chance that we are introducing errors into the table. A better approach would be to add a data pipeline to download these files from EIA, transform them, and apply standardized, replicable, and transparent rules for implicit data types. 

We may also need to look at emission factors from EPA since there are some factors that do not appear in the EIA tables.

Issue #218, opened 2022-09-06
Landfill Gas (LFG) SO2 emissions factors


grgmiller:
Related to LFG emissions factors and issue #69 and #150: eGRID uses two sources of data for their SO2 emisisons factors: EPA's AP-42 document, and the EIA electric power annual (which also uses AP-42 as a source). 

EIA assumes that LFG has the same SO2 emissions factor as natural gas (0.0006 lb/thousand ft3), but eGRID uses an SO2 emissions factor for LFG of  0.045 lb/mmbtu, which are calculated based on the 1995 version of AP-42. However, I think that this emission factor may be out of date and overestimating LFG emissions. For example, based on the EIA emission factor, the average SO2 emission factor for all LFG generators in 2020 is about 0.003 lb/mmbtu. 

I also tried to slog through the calculation that would have been used to calculate the default emission factor from AP-42, but using the updated draft version of section 2.4 published in 2008. I definitely had to brush off some chemistry knowledge, so the following likely needs to be double checked, but the default number I arrived at was 0.025 lbSO2/mmbtu (which is still way less than the default eGRID value). 
![image](https://user-images.githubusercontent.com/45949268/188521817-1fc8cd22-ba71-455d-ac9b-8bfbe15382bd.png)

Note that the concentration of dimethyl sulfide is highlighted in red because it is unclear in the draft whether the value should be 5.66 or 0.566. The value listed in table 2.4-1 is 5.66E+0, but the lower value of the default C_S mentioned at the top of page 2.4-11 is 33 ppmv, which only works if the concentration of dimethyl sulfide is 5.66E-1. 

Note that if we use the default lower value of 33 ppmv S (for newer landfills since 1992) the EF is 0.022 lb SO2/mmbtu, and if we use the higher value of 47 ppmv S (for older landfills pre 1992) the EF is 0.031 lb SO2/mmbtu).

**References**
- 2008 draft version of AP-42 Section 2.4: https://www.epa.gov/sites/default/files/2020-10/documents/d02s04_0.pdf
- 1998 (Fifth Edition) version of AP-42 section 2.4: https://www.epa.gov/sites/default/files/2020-10/documents/c02s04.pdf

grgmiller:
It looks like this issue is causing a large discrepancy in SO2 emissions between OGE and eGRID.

Issue #150, opened 2022-07-22
Assumed NOx/SO2 emission factors


grgmiller:
For its table of boiler type-specific NOx emissions factors, eGRID uses this table from EIA for emissions factors: https://www.eia.gov/electricity/annual/html/epa_a_02.html. 

However, where there is not a specific NOx emission factor in this table (for example for ST prime movers with SUB fuel, but no information about the boiler type, it appears that eGRID assumes the lowest emission factor of all emission factors for this fuel type, rather than using the lower emission factor. In this case the factors for SUB range from 5-24, and eGRID uses 5.

On one hand, this could be considered conservative (ie we know that there is at least this much emissions), but on the other hand, this could be under-counting NOx emissions.

Issue #69, opened 2022-06-08
Look into updated default emission factors


grgmiller:
The eGRID2020 technical guide notes that:
> The emission factors are primarily from the default CO2 emission factors from the EPA Mandatory Reporting of Greenhouse Gases Final Rule (EPA, 2009, Table C-1). For fuel types that are included in eGRID2020 but are not in the EPA Mandatory Reporting of Greenhouse Gases Final Rule, additional emission factors are used from the 2006 Intergovernmental Panel on Climate Change (IPCC) Guidelines for National Greenhouse Gas Inventories and the EPA Inventory of U.S. Greenhouse Gas Emissions and Sinks: 1990-2015 (IPCC, 2007a; EPA, 2017).

However, it is unclear whether there might be more up to date emissions factors that should be used:
- [ ] Look into whether there are newer default GHG emission factors than those in the 2009 EPA Final Rule
- [ ] Look into whether IPCC emission factors have been updated since 2006
- [ ] Look into whether there are updates to the 2017 EPA Inventory of U.S. Greenhouse Gas Emissions and Sinks

The technical guide also notes:
> Several fuel types do not have direct reported emission factors, so emission factors from similar fuel types are used:
  • The emission factor for natural gas is used to estimate emissions from process gas and other gas;
  • The emission factor for anthracite, bituminous, and lignite coal are used to estimate emissions from refined coal and waste coal; and
  • The emission factor for other biomass liquids is used to estimate emissions from sludge waste and liquid wood waste

- [ ] Research whether there are now specific emission factors for process gas (PRG)
- [ ] Research whether there are now specific emission factors for other gas (OG)
- [ ] Research whether there are now specific emission factors for refined coal (RC)
- [ ] Research whether there are now specific emission factors for waste coal (WC)
- [ ] Research whether there are now specific emission factors for sludge waste (SLW)
- [ ] Research whether there are now specific emission factors for liquid wood waste (WDL)

grgmiller:
IPCC Factors come from Chapter 2: https://www.ipcc-nggip.iges.or.jp/public/2006gl/pdf/2_Volume2/V2_2_Ch2_Stationary_Combustion.pdf, Table 2.2
To convert from (kg of greenhouse gas per TJ on a Net Calorific Basis) to lb/mmbtu, you must multiply by 0.002326

grgmiller:
[...]

PR #373, opened 2024-06-27
Update Uncontrolled NOx and SO2 factors

grgmiller:
### Purpose
This PR was started to add NOx and SO2 factors for boiler configurations in the historical data that were not previously specified. This is part of the effort to improve NOx and SO2 data quality for new historical years from 2005-2018.

**Programmatic creation of the factor tables**
Instead of continuing to manually update these in the table, we add a script to programmatically create the nox and so2 factor reference tables to address https://github.com/singularity-energy/open-grid-emissions/issues/276. This is meant to reduce the risk of typos, ensure consistency in the methodology, and also add transparency to how these factors are created.

**Fluidized Bed Boilers**
As noted by the EIA's Electric Power Annual [technical notes](https://www.eia.gov/electricity/annual/pdf/tech_notes.pdf), while the uncontrolled emissions factors published by EIA need to be adjusted to account for each plant's operational pollution control equipment, the exception is for fluidized bed boilers: 
>A special case for removal of SO2 is the fluidized bed boiler, in which the sulfur removal process is integral with the operation of the boiler. The SO2 emission factors shown in Table A.1. for fluidized bed boilers already account for 90 percent removal of SO2 since, in effect, the plant has no uncontrolled emissions of this pollutant.

Previously, we had been applying pollution control efficiency values to adjust so2 emissions from these boilers, which would underestimate emissions. We now do not adjust so2 emissions for this type of boiler since the calculated values already represent controlled emissions. It looks like this may have only affected a small number of plants, since many plants did not report control efficiencies for fluidized bed technologies in EIA-923.

**Update default assumed value**
This PR also includes an important methodological update on the assumptions used for assigning NOx and SO2 factors for boilers with missing boiler configuration data. When boiler configuration data is unknown, we had previously been using the lowest fuel-specific NOx or SO2 factor of any boiler firing type, with the reasoning that this is the minimum amount of emissions that could be expected. This was also the approach that eGRID had used in previous years (looking at more recent versions of eGRID, it looks like they may have abandoned this approach). 

However, there are several drawbacks of this approach. For SO2, since the fluidized bed factors are already controlled, this approach would mean generally assuming that the generator is a fluidized bed boiler, which would be applying a special case as the default backstop. In the case of NOx, this could lead to the assumptions of different types of boilers being the standard default. The EIA's default assumption, in the case that boiler information is unavailable, is to assume that the boiler has a dry-bottom firing method falling into the "all other category." We adopt this assumption in this PR, which generally have the effect of increasing NOx and SO2 emissions in these cases.
https://github.com/singularity-energy/open-grid-emissions/issues/150

### What the code is doing
How is the purpose executed?

### Testing
In the notebook, I test for:
- duplicate factors
- How much the current factors differ from the factors already in `reference_tables`
- That all historical boiler configurations are covered.

Ran pipeline for 2017-2018

### Where to look
- src/oge/create_nox_so2_factors.py

### Usage Example/Visuals
For NOx, there are 80 factors that are different from previous.
image

For SO2, there are 288 factors that are different from the previous factors.
image

Inspecting these differences reveals that they are based on the updated default assumption, and how we treat combustion turbines with a specified boiler firing type. Where previously we had been assigning the combustion turbine factor, we now use the factor associated with the boiler firing type, since these may be associated with HRSG configurations (see: https://www.eia.gov/todayinenergy/detail.php?id=52778)

There are a few configurations for which NOx factors are still missing, but these seem to be anomalies (eg a hydro plant burning oil). This may just be a figment of the test logic I set up and these anomalous configurations may not appear in the real data.
image

### Review estimate
20-30 min

### Future work
Updating LFG factors: https://github.com/singularity-energy/open-grid-emissions/issues/218
Default factor assumptions: https://github.com/singularity-energy/open-grid-emissions/issues/69

### Checklist


src/oge/load_data.py
```diff
@@ -1224,6 +1202,9 @@ def load_emissions_controls_eia923(year: int) -> pd.DataFrame:
         emissions_controls_eia923["report_date"] = emissions_controls_eia923[
             "report_date"
         ].astype("datetime64[s]")
+        emissions_controls_eia923["equipment_tech_description"] = (
+            emissions_controls_eia923["equipment_tech_description"].str.lower()
+        )
     else:
         logger.warning(
             "Emissions control data prior to 2012 has not been integrated into the data pipeline."
```
rouille, line 1210: Is that 2012 or 2008?

src/oge/create_nox_so2_factors.py
```diff
@@ -0,0 +1,574 @@
...
+    There are several filling steps:
+    1. Missing energy_source_codes: The EIA tables do not provide NOx or SO2 factors for
+    MSB, MSN, or SC fuel types. For MSB and MSN (the biogenic and nonbiogenic
+    of MSW), we set the factors equal to the specified MSW factor. For SC (Coal Synfuel)
+    We set the factor equal to the refined coal (RC) factor, which itself is assumed to
+    be equal to the bituminous coal (BIT) factor.
+    2. Factors for units missing boiler_firing_type (BFT) and wet_dry_bottom (WDB) data:
+    According to the technical appendix for the EIA's Electric Power Annual (see:
+    https://www.eia.gov/electricity/annual/pdf/tech_notes.pdf), whenever a unit is
+    missing BFT and WDB data (in the case of NOx), they apply they assume it is a dry-
```
rouille, line 313: they apply they assume -> they assume

src/oge/create_nox_so2_factors.py
```diff
@@ -0,0 +1,574 @@
...
+            "data_source",
+            "unit",
+            "cyclone_firing",
+            "fluidized_bed_firing",
+            "stoker",
+            "tangential_firing_dry",
+            "tangential_firing_wet",
+            "other_dry",
+            "other_wet",
+            "GT",
```
rouille, line 51: Should it be `CT` for combustion turbine? Just want to make sure it is not a typo

src/oge/create_nox_so2_factors.py
```diff
@@ -0,0 +1,574 @@
...
+
+    factor_columns = [
+        "cyclone_firing",
+        "fluidized_bed_firing",
+        "stoker",
+        "tangential_firing_dry",
+        "tangential_firing_wet",
+        "other_dry",
+        "other_wet",
+        "GT",
```
rouille, line 65: Same comment here `GT` -> `CT`

src/oge/create_nox_so2_factors.py
grgmiller: If you look at the prime mover metadata (https://catalystcoop-pudl.readthedocs.io/en/latest/data_dictionaries/codes_and_labels.html#core-eia-codes-prime-movers), GT is the code for a general combustion turbine, while CT is specifically a combustion turbine that is part of a combined cycle unit. 
We end up applying this EF to the CT prime mover later in the pipeline.

src/oge/create_nox_so2_factors.py
grgmiller: See above comment

src/oge/create_nox_so2_factors.py
grgmiller: Fixed

src/oge/load_data.py
grgmiller: PUDL has integrated data from 2008 onward, but their version of the table does not currently include columns for hours_in_service and ozone_season_nox_emission_rate. See: https://github.com/catalyst-cooperative/pudl/issues/3689. 

We are still manually loading the EIA-923 table from a raw download, and we have only integrated back to 2012 so far.

src/oge/create_nox_so2_factors.py
```diff
@@ -0,0 +1,574 @@
...
+    # apply the combustion turbine factor to all combustion-turbine-type prime movers,
+    # including those at combined cycle plants, and CAES plants
+    for pm in ["CA", "CC", "CS", "CT", "CE"]:
+        df[pm] = df["GT"]
+
+    # where we have a prime mover with no BFT or WDB information, we assume a dry-bottom
+    # boiler in the "all other" BFT category. This is consistent with the assumption made
+    # by the EIA. see: https://www.eia.gov/electricity/annual/pdf/tech_notes.pdf
+    # for pm in ["ST","OT"]:
+    #    df[pm] = df[default_value]
```
rouille, line 374: Why is that commented? Don't we want to assign factors for these PMs?

rouille (approved):
Thanks

PR #374, opened 2024-06-29
Update Energy Source Codes and Utility Name Map

grgmiller:
### Purpose
Update some of the reference tables, specifically:
- Add SG to `energy_source_groups` and update description of SC.
- Update the `utility_name_ba_code_map` with additional maps.
- Updated the notebook used to manually identify OTH energy source codes, but did not update the table, as this is more involved and would take longer to implement.

Closes CAR-4380 and CAR-4381

### What the code is doing
How is the purpose executed?

### Testing
Ran pipeline for 2012

### Review estimate
5 - 10 min

### Future work
Update the OTH energy source codes. One thing I realized was that these can change over time for a specific plant, so we may want to re-think how this table works in the future. We may need to try to identify fuels based on the heat content, or assign years to the table.

### Checklist


rouille (approved):
Thanks

PR #378, opened 2024-07-09
Add geopy to pyproject dependencies

grgmiller:
### Purpose
In https://github.com/singularity-energy/open-grid-emissions/pull/368 we added `geopy` as a dependency to OGE, but this dependency was not updated in the `pyproject.toml` file, which means that certain modules are unusable if imported into another repo. This PR adds this dependency.


rouille (approved):
Thanks!

PR #379, opened 2024-07-15
Compare plants coordinates from PUDL and EIA-860

rouille:
### Purpose
Compare plants coordinates from PUDL and EIA-860. Closes CAR-4443. 

Outputs from notebook will be cleared after review

### What the code is doing
* Load plants entity in PUDL and EIA-860
* Get plants available in both dataset
* Calculate for each plant the angular distance between its coordinates from PUDL and EIA-860

### Testing
N/A

### Where to look
Scroll through the notebook

### Usage Example/Visuals
image

### Review estimate
5min

### Future work
Look at plants with large angular separation

### Checklist

### DIFF
--- a/notebooks/explore_data/compare_plant_coordinates.ipynb
+++ b/notebooks/explore_data/compare_plant_coordinates.ipynb
@@ -1,0 +1,231 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Compare Plant Coordinates from PUDL and EIA-860 "
+   ]
+  },
+  {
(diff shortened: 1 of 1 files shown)


grgmiller (approved):
Looks good as an initial exploration. 
I'll note that when I was exploring some GRETA updates, a lot of plants got updated timezones based on the coordinate interpolation, and I'm not sure if they are all correct. A lot of them were tricky, like hydro plants on a river, whose official address is on one side of the river (in one time zone), but the lat/long is in the middle of the river (in a different time zone). We may need to do some additional validation here in the next round of improvements.

PR #382, opened 2024-07-24
Discard non-operational generators when calculating plant capacity

rouille:
### Purpose
Fix bug where nameplate capacity of non-operational generators were considered in the calculation of the nameplate capacity. Closes CAR-4479

### What the code is doing
Add a condition on the operational status.

### Testing
Compared nameplate capacity of plants 113, 117 and 2442 that were flagged with a wrong nameplate capacity before changes. 

### Where to look
Very few changes.

### Usage Example/Visuals
N/A

### Review estimate
2min

### Future work
N/A

### Checklist


grgmiller (approved):
Looks good, thank you!

PR #383, opened 2024-07-26
Revert removal of GTN shift factors

grgmiller:
### Purpose
In https://github.com/singularity-energy/open-grid-emissions/pull/370, I had removed the shift factors used for gross to net generation conversions without fully testing the consequences, but this resulted in a large number of generators having net generation that did not match the net generation in EIA-923. At the time, this seemed ok, but since then through some conversations with someone at EIA, it seems the best course to trust the EIA data by default and continue our existing methodology until we have time to do a deeper dive into this.

Closes CAR-4490

### What the code is doing
Move the subplant shift factors higher in the GTN hierarchy, so that they are applied before the fleet average ratio.

Also removes some outputs from the sandbox notebook that we hadn't previously deleted from a previous PR.

### Testing
Ran the pipeline for 2022, and we are back to only having the original 5 plants with mismatched net generation:
image

Also, re-ran 2005 pipeline. We went from 100 plants with issues to only 22.

### Review estimate
5 minutes. This just re-implements the GTN hierarchy as it existed prior to #370 while keeping the other fixes introduced in that PR.

### Future work
None

### Checklist


rouille (approved):
Sounds good

PR #385, opened 2024-07-31
Fix missing capacity in plant static attributes

grgmiller:
### Purpose
The plant static atttributes file included missing values for the plant capacity. This PR fixes that. 

### What the code is doing
This includes several fixes:
- forward fill the operational status column so we don't drop generators with a missing status
- If a generator retires partway through a year, we want to keep that capacity, so change status to "existing"

Includes multiple steps for ensuring complete data for capacity. These values will
    only be used if there is no "existing" capcity reported for a year.
    1. Some plants report data before reporting to EIA-860. In this case, find the
        earliest reported capacity for the plant and fill missing capacity values.
    2. Sometimes plants report data before they are operational. Find the capacity of
        any proposed generators and use this to fill missing capacity values.
    3. Sometimes plants report data after they have retired. In this case, find the
        capacity of all generators in the latest year in which the plant was operating.

Also added a validation check to look for missing values in all columns of the plant attributes table except for the plant retirement date, which we expect to have missing values.

### Testing
Re-running 2005-2022.

Example of new error message for plant static attributes:
image

### Where to look
look in helpers

### Review estimate
10 min

### Future work
- Fill in other missing values flagged in plant static attributes
- In many instances I'm getting a connection error with nominatim - we may want to set up some sort of handling for this:
```
2024-07-31 00:35:41,038 [INFO] oge.data_pipeline:251 5. Loading plant static attributes
2024-07-31 00:36:08,477 [WARNING] urllib3.connectionpool:826 Retrying (Retry(total=1, connect=None, read=None, redirect=None, status=None)) after connection broken by 'ReadTimeoutError("HTTPSConnectionPool(host='nominatim.openstreetmap.org', port=443): Read timed out. (read timeout=1)")': /reverse?lat=29.867431&lon=-93.965775&format=json&addressdetails=1

```

### Checklist


rouille (approved):
Thanks!

PR #384, opened 2024-07-26
Update to 0.5.0 and change s3 directory

grgmiller:
### Purpose
Updates the version to 0.5.0. 
In s3, we will also begin saving versioned OGE data for the current and most recent data version. This data will now be saved in a directory with the version number, like `s3://open-grid-emissions/open_grid_emissions_data/v0.5.0`. Thus, I updated the filepath handler to use this convention for the code going forward.

I also updated the pipfile.lock to address all of the dependabot issues.

Updates the conda environment.yml file

Closes CAR-4491, CAR-4451, CAR-4466, CAR-4439

### Testing
Points to correct location:
image

### Review estimate
5 min

### Checklist
- [ ] Update the documentation to reflect changes made in this PR
- [ ] Format all updated python files using `black`
- [ ] Clear outputs from all notebooks modified
- [ ] Add docstrings and type hints to any new functions created

PR #372, opened 2024-06-26
Update manual epa eia crosswalk reference table

rouille:
### Purpose
A CEMS only with missing fuel category has been identified in EIA with a different identification number. Matching has been done using the plant name, the plant location (state name and coordinates) and comparing generation id vs emission unit id 

### What the code is doing
N/A

### Testing
Ran pipeline for 2005.

### Where to look
The **src/oge/reference_tables/epa_eia_crosswalk_manual.csv** file

### Usage Example/Visuals
N/A

### Review estimate
2min

### Future work
N/A

### Checklist
- [x] Update the documentation to reflect changes made in this PR
- [x] Format all updated python files using `black`
- [x] Clear outputs from all notebooks modified
- [x] Add docstrings and type hints to any new functions created

PR #362, opened 2024-05-17
Add 2005, 2006 and 2007 years

rouille:
### Purpose
Allow to run pipeline for years 2005, 2006 and 2007. Closes CAR-4211.

Add missing typehints and add/format docstrings in all functions located in the `load_data` and `download_data` modules.

### What the code is doing
* Print a warning message and return `download_data.download_raw_eia923` instead of raising an error for years < 2008
* Updates reference tables to add the `SC` energy source code that is used for years < 2008 in place of `SGC`

### Testing
Run 2005, 2006 and 2007 pipelines successfully.

### Where to look
* the `download_data.download_raw_eia923` where we warn the file is not available and return
* the `load_data.load_pudl_table` where a warning is added if the table is empty
* the 2 reference tables (CSV files) where we added information for the `SC` energy code. 

### Usage Example/Visuals
From log file:
```
2024-05-17 11:21:37,194 [WARNING] oge.oge.download_data:313 EIA-923 data is unavailable for 2006.
...
2024-05-17 11:22:32,070 [INFO] oge.data_pipeline:169 3. Cleaning EIA-923 data
2024-05-17 11:22:33,969 [WARNING] oge.oge.load_data:670 denorm_boiler_fuel_monthly_eia923 is empty
2024-05-17 11:22:34,136 [WARNING] oge.oge.load_data:670 denorm_generation_monthly_eia923 is empty
2024-05-17 11:22:34,660 [WARNING] oge.oge.load_data:670 boiler_generator_assn_eia860 is empty
```

### Review estimate
2 min if you don't pay attention to added/formatted docstrings. 20min otherwise

### Future work
Add information in the plant attributes file.

### Checklist
- [ ] Update the documentation to reflect changes made in this PR
The README will be added in a separate PR before merging the `historical_coverage_feature` branch into `development`
- [x] Format all updated python files using `black`
- [x] Clear outputs from all notebooks modified
- [x] Add docstrings and type hints to any new functions created

PR #361, opened 2024-05-16
Remove add_subplant_id optional argument

rouille:
### Purpose
Clean and allocate generation and fuel in EIA-923 data at sub-plant level only. The option to do it at the plant level was bugged (see #360) and is now deactivated. Closes #360. Closes CAR-4207.

Add missing typehints and docstrings to the functions 

### What the code is doing
Remove `add_subplant_id` from functions and discard `if` logic. 

### Testing
Run 2012 pipeline successfully

### Where to look
The `oge.data_cleaning.clean_eia923` and `oge.data.create_primary_fuel_table` are the 2 functions that are modified.
Add missing typehints and docstrings to these functions.

### Usage Example/Visuals
N/A

### Review estimate
5min

### Future work
N/A

### Checklist
- [x] Update the documentation to reflect changes made in this PR
- [x] Format all updated python files using `black`
- [x] Clear outputs from all notebooks modified
- [x] Add docstrings and type hints to any new functions created

PR #386, opened 2024-07-31
Historical coverage feature / v0.5.0

grgmiller:
### Purpose
This PR merges several months of work to expand the historical coverage of the data back to 2005. All of this work has been previously reviewed. 

**NOTE: I am proposing that we merge directly into `main` and then rebase development to match main once v0.5.0 is released.**

**Draft release notes**
Expands historical coverage of OGE to include monthly and annual data for 2005-2018 (#295 and #362)

In addition to the new data, users should expect changes to the existing 2019-2022 data: NOx and SO2 totals may change for some plants, net generation totals may change for some plants, data may change for CHP plants (see the "methodological updates" section for more details)

Inputs:
- Updates to use the most recent data version of PUDL (v2024.5.0). This includes a re-release of the 2022 EIA-923 data, which may change some of the 2022 results. 
- Updates reference tables including the `energy_source_groups` file, and the `utility_name_ba_code_map` file (#374), and `epa_eia_crosswalk_manual` (#372), and `emission_factors_for_co2_ch4_n2o` (#377)

Outputs
- All output files (those in the `outputs/` directory are now saved as compressed `.csv.zip` files instead of `.csv` files. This reduces the disk space of the outputs folder from approximately 16GB to 2.5GB. (#366)
- Expands the data in the plant_static_attributes table to include location data (lat/long, address) and nameplate capacity (#364, #382, #385); commercial operation dates and retirement dates (#367). We also screen for and correct erroneous lat/long data (#368)
- Fixes a bug where the "total" values in the `outputs/annual_generation_averages_by_fuel` file were not being calculated correctly

Methodological updates
- When calculating the electric allocation factors for combined heat and power (CHP) plants, we previously were calculating this at the generator level, which was introducing bugs for certain combined cycle units when fuel and generation is reported for different generators at the same subplant. We now calculate this factor at the subplant level (#363)
- Fixes several bugs with the gross-to-net generation conversions where anomalous fleet-average ratios were being introduced, and default factors were not being mapped to certain generators. Also fixed a bug where GTN ratios were being calculated where there was missing gross generation or net generation data. (#370, #375, #383)
- Updates uncontrolled NOx and SO2 factors to align assumptions with those used by the EIA Electric Power Annual, and to fix a bug where we were adjusting the SO2 values for fluidized bed boilers, even though the control efficiencies are already incorporated into the uncontrolled emission factors (#373). In addition, because fuel sulfur content data is not available pre-2008, we use sulfur content values averaged from 2008-2012 to backfill the missing data. When calculating backstop values for missing values in any year, we now use state-specific values (rather than national-average) to reflect differences in the sulfur contents of fuels being delivered in specific parts of the country (#376)

Other fixes
- Remove the option to run the EIA-923 allocation at the plant level. This was an artifact that was no longer used (#361)
- Clean up function typehints and continue converting docstrings to Google format
- Updates where files are stored and accessed from in s3 (#384)


rouille (approved):
Thanks. In the methodological changes/updtaes section of the PR description (or in the release notes), we can add the updates to the calculation of the sulfur content (national to state level backstop) and the filling of missing values for years 2005-2007 (use 2008-2012 data)

PR #389, opened 2024-08-09
Update EPA/EIA crosswalk of plant 55641

grgmiller:
### Purpose
Generators CTG3, CTG4, STG2, and PV1 were proposed to be added to the Riverside Energy Center (Plant 55641) and were in testing as of 2019 with scheduled COD in 2020. However, starting in the 2020 EIA data, these generators disappeared from this plant, and reappeared as the West Riverside Energy Center (Plant 64020). Closes CAR-4543.

2019 Proposed Generator Data:
image

2020 Operational Generator Data:
image

I've updated the manual EPA-EIA crosswalk file to represent this. STG2 does not appear to directly show up in the data

### Testing
We would expect the subplant crosswalk to reflect the new mapping.

As expected, the EPA 55641 is no longer associated with those two units at EIA 55641 (although they still appear in the crosswalk, which is strange since they were never built)
image
These two units are now associated with plant 64020, and are correctly linked to the same subplant as STG2
image

### Review estimate
5 min

### Future work

### Checklist


rouille (approved):

grgmiller:
@rouille you're supposed to be on vacation!

gailin-p (approved):
LGTM, I assume all the other line changes are line ending? Might be nice to switch to unix line endings here too

PR #393, opened 2024-11-13
Update year to 2023

grgmiller:
### Purpose
This is a small PR that just does the minimum updates to download 2023 data and run the pipeline, without actually fixing any data quality issues yet. The pipeline did successfully complete but there are a number or errors raised, including a large amount of negative fuel in MISO. It is possible that these will be fixed once we properly go through the entire update process.
Advances CAR-4691

### What the code is doing
How is the purpose executed?

### Testing
Ran the pipeline

### Where to look

### Usage Example/Visuals

### Review estimate
5 minutes

### Future work
The data pipeline log file was over 300MB because it output a large amount of hourly data in the warning. We should limit these warnings in the future to just print a sample of the data

### Checklist

### DIFF
--- a/src/oge/constants.py
+++ b/src/oge/constants.py
@@ -11,7 +11,7 @@ earliest_validated_year = 2005
 # profiles
 earliest_hourly_data_year = 2019
 # latest_validated_year is the most recent year for which OGE data has been published
-latest_validated_year = 2022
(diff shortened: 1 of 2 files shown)


rouille (approved):
Thanks

PR #401, opened 2024-12-13
Refactor function writing power sector data

rouille:
### Purpose
Monthly and annual power sector data were written twice. Once in Step 12 and one in step 18. This is because the monthly and annual data are always written when the `write_power_sector_results`, hourly data are written or not depending on the setting of a boolean parameter named `include_hourly`. We use the same behavior for monthly and annual.

Also, add CPLW to `BA_930_INCONSISTENCY` dictionary.

Advances CAR-4691

### What the code is doing
Add the `include_monthly` and `include_annual` parameters to the `write_power_sector_results` function and use `if`/`else` logic to write or not data.

### Testing
Ran the pipeline for 2023

### Where to look
* the `write_power_sector_results` function in the `oge.output_data` module and the pipeline where this function is called
* the updated `BA_930_INCONSISTENCY` dictionary where CPLW has been added for 2023

### Usage Example/Visuals
N/A

### Review estimate
5 min

### Future work
Locate subplant/plant(s) in PJM introducing missing fuel category for power sector data

### Checklist

### DIFF
--- a/src/oge/output_data.py
+++ b/src/oge/output_data.py
@@ -556,6 +556,8 @@ def write_power_sector_results(
     path_prefix: str,
     skip_outputs: bool,
     include_hourly: bool,
+    include_monthly: bool,
+    include_annual: bool,
 ):
     """Helper function to write combined data by BA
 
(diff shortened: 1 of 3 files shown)


src/oge/output_data.py
```diff
@@ -669,47 +672,31 @@ def write_power_sector_results(
...
-                year,
-                ba,
-                "power_sector_data/annual/",
-                path_prefix,
-                skip_outputs,
-            )
+                if include_annual:
+                    agg["annual"] = ["fuel_category"]
+
+                for k, v in agg.items():
```
grgmiller, line 682: instead of k,v can we make these descriptive variables like agg_level, groupby_cols

grgmiller (approved):
One small change requested but otherwise looks like an elegant way to refactor the code.

PR #402, opened 2024-12-14
Strip leading zeros from CEMS `emission_unit_id_epa`

grgmiller:
### Purpose
We had noticed that in 2023, the PJM power sector results file had a missing fuel category. I traced this to plant 50852, and it turns out the cause of the issue was that the fuel type backfill using the PSDC was failing because we were not stripping leading zeros from the unit ID when loading the table, so the merge was not working. 

After digging into this more, I realized that pudl does not automatically strip leading zeros from `emissions_unit_id_epa` (see https://github.com/catalyst-cooperative/pudl/issues/3992). This PR adds functionality to always strip these leading zeros whenever loading CEMS data

This PR also updates the sandbox notebook to use 2023 as the year, and also adds some starter functions for loading intermediate output data for exploration since these frequently need to be loaded.

### Testing
Ran pipeline for 2023

### Where to look

### Usage Example/Visuals

### Review estimate

### Future work

### Checklist


rouille (approved):
Thanks

PR #403, opened 2024-12-20
Expand eia-epa crosswalk

rouille:
### Purpose
In order to avoid missing fuel category in the monthly and annual power sector data files, we:
* Drop subplants that have zero fuel consumption and zero generation when aggregating to fleet
* Expand EIA-EPA crosswalk and temporary primary fuel tables  
* Identify generators that suddently disappears but are not marked as retired 

Advances CAR-4691

### What the code is doing
the generators disappearing from the data from one year to the other are added to the list of retired generators complete coverage when a generator continues reporting CEMS data even after retiring.

### Testing
Ran pipeline for 2005 and 2008

### Where to look
* the manual tables
* and the logic in the `add_recently_retired_generator_ids_to_df` function of the `oge.data_cleaning` module

### Usage Example/Visuals
N/A

### Review estimate
10min

### Future work
Create a primary fuel attributes table for subplants

### Checklist


grgmiller (approved):
Looks good to me, except I think the dropping zero code should go in a different function. I made that change in #405, so let's merge 405 into here before merging this PR into dev

PR #407, opened 2024-12-20
Final Cleanup

grgmiller:
### Purpose
Final cleanup before the 2023 data release, including various small fixes: 
- bump the version to 0.6.0 
- Update the pipfile.lock to address security warnings
- Add handling for geopy timeout errors: when using geopy to lookup missing lat/long or city/state/county, sometimes the request times out or the geocoder is unavailable. Previously this was causing the entire pipeline to crash. This adds a retry loop to retry the request once if unsuccessful, as well as handling both errors and returning nans if it fails twice. 
- Fixes a bug that introduced duplicate columns in the plant_data output file. 
- Fixes a bug where the `downloads/eia923` directory was not being created if the year was < 2008 and the directory did not already exist
- For years prior to 2019, outputs the `plant_static_attributes` table prior to the `plant_data` results instead of after so that it is available for debugging if the `plant_data` export fails
- Updates the "H" frequency to "h" to address a deprecation warning.
- Re-arranges the column order in the annual plant data file to include the plant attributes after the plant ID and before the quantitative data.

Advances CAR-4691

### Testing
I'm going to do a full run of 2005-2023 to make sure everything is working, then merge

### Review estimate
5 min

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -22,6 +22,6 @@ authors:
 identifiers:
   - type: doi
     value: 'https://zenodo.org/doi/10.5281/zenodo.7062459'
-version: 0.5.0
+version: 0.6.0
 license: MIT
-date-released: '2024-08-01'
+date-released: '2024-12-21'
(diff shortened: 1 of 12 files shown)


rouille (approved):
🚀

PR #409, opened 2025-07-16
Fix circular imports

grgmiller:
### Purpose
Upstream repos using OGE were failing pytest due to circular imports in OGE between the `helpers` and `validation` module. This PR restructures the repo to avoid these circular imports, mostly by removing any calls to `validation` from the `helpers` module. Also fixes a circular import between `load_data` and `validation`.

Also performs some additional cleanup:
- Updates package dependencies to address security warnings
- Updates formatting using latest version of `ruff`

### What the code is doing
- Moves `identify_percent_of_data_by_input_source` from `validation` to `output_data`
- Moves some validation functions from helpers to the main data_pipeline
- Moves `test_for_missing_subplant_id` from validation to helpers since it is only used by a single function in helpers.

### Testing
Asked cursor to analyze for circular imports and it could not find any more.

### Review estimate
10 minutes

### Future work
Implement test for circular imports?

### Checklist

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -22,6 +22,6 @@ authors:
 identifiers:
   - type: doi
     value: 'https://zenodo.org/doi/10.5281/zenodo.7062459'
-version: 0.6.0
+version: 0.6.1
 license: MIT
-date-released: '2024-12-24'
+date-released: '2025-09-26'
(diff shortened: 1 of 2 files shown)


grgmiller:
I'd propose that perhaps we do this as a patch release if this is causing issues elsewhere

rouille:
> I'd propose that perhaps we merge directly into main and do this as a patch release if this is causing issues elsewhere. If you agree I can update the version number before merging

Sounds good

rouille (approved):
Thanks

PR #419, opened 2025-11-25
Improve pipeline memory use

grgmiller:
### Purpose
While running the pipeline locally, I keep running into memory errors where my computer runs out of RAM while running the pipeline. This PR attempts to use memory more efficiently to prevent these memory errors from happening. 

To do this, we utilize several strategies:
- Explicitly `del` dataframes from the pipeline once we are done using them and also periodically run garbage collection using `gc`
- When dropping columns, drop inplace rather than creating a copy (`df = df.drop(column="columnname")`
- When concatenating files, where appropriate, set `copt=False` to prevent creating unnecessary copies.
- When creating new copies of files, copy after filtering (`df[filter].copy()` rather than `df.copy()[filter]`
- Drop steam load and emissions mass measurement code columns from the CEMS data after initial data cleaning since they are unused. 

### Testing
Ran the pipeline

### Review estimate
5-10 min

### Future work
I'm still unable to get through the pipeline without crashing my memory, although I can now get much further. To ultimately fix this, I think one of the biggest memory hogs is that we are keeping a large amount of zero data in the CEMS dataframe and moving/copying that around. We used to remove data for months with zero data for a unit `remove_cems_with_zero_monthly_data()` but we stopped using that in 2023 due to some downstream issues that it was causing (see https://github.com/singularity-energy/open-grid-emissions/pull/322). However, we may just want to address that issue directly and/or figure out an alternative way to use zero data more efficiently. 

### Checklist

### DIFF
--- a/src/oge/reference_tables/ba_reference.csv
+++ b/src/oge/reference_tables/ba_reference.csv
@@ -25,190 +25,192 @@ BCA,Batesville Control Area,,,,Yes,1/1/2004,9/30/2006,,,,FERC,23
 BCHA,British Columbia Hydro and Power Authority,,Canada/Pacific,Canada/Pacific,No,1/1/2004,,,,,EIA & FERC,24
 BCTC,British Columbia Transmission Corporation,,,,No,10/1/2008,4/1/2015,,,,FERC,25
 BERC,"DECA, LLC",,,,Yes,1/1/2004,7/1/2006,,,,FERC,26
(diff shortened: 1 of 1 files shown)


rouille (approved):
LGTM. We should consider in the `column_checks` module set some of the column to `float32` in place of `float64`. I would say this could have a significant effect on memory usage.

PR #421, opened 2025-12-03
Miscellaneous cleanup

grgmiller:
### Purpose
In this branch, we wanted to explore if we could optimize memory usage in the pipeline by dropping months of CEMS data with all zero values. We did not end up implementing this because it would lead to downstream issues (see notes in comments below). However, we did make several other changes.  

We previously had a note that dropping months of CEMS data with all zeros could cause problems in the pipeline. After looking into this, I think this is because:
- If we are missing CEMS data, then we automatically use EIA-923 for that month. 
- However, some generators report annual data to EIA-923, which EIA-923 allocates out to months using a non-precise method. 
- This means that if we are missing CEMs data for a month, and EIA has mistakenly allocated generation to that month, we are going to mis-allocate or even potentially under/over count emissions data. If we keep the zeros, then this cannot happen. 

For now, I'm adding the reporting frequency column to some of our outputs to evaluate if this is part of the problem.

Other changes:
- When calculating hourly profiles for shaping, we now ignore CEMS profiles if any of the values are negative. CEMS is used only as a backstop, so if there are negative CEMS values, this will get translated through to the hourly profile and cause potentially negative emissions issues. We drop these values, so now the pipeline should default to a flat profile. This fixes the issue that we were seeing with SWPP petroleum data. 
- Remove the dask dependency. There was a conflict between Python 3.11.9 and The pinned dask version. I think the original reason we pinned dask has now been fixed by more recent versions. 
- Adds error handling for running the pipeline without an internet connection. The reverse coordinate lookup will not work, but this allows for offline testing.
- Removes the `--small` argument from the code
- Fixes bug with EIA-923 allocation validation check that was not reflecting all input data.

### What the code is doing
How is the purpose executed?

### Testing
Ran pipeline. I'm still running into memory issues.

### Usage Example/Visuals

### Review estimate

### Future work

### Checklist

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -246,11 +246,11 @@
         },
         "arelle-release": {
             "hashes": [
-                "sha256:0f8376121d08468f6ddda59f60c62dfbb91b63a92da6c5aa8bbcd6676b1fbef9",
-                "sha256:7202afc1ac75bc3302bc77644710fdc018ea904954e1779bfca46e10977decb6"
+                "sha256:7b9df61893f02ca5e3b4d1e696d275905ecc5f3dd2df711d30e15236a8336af1",
+                "sha256:85fd4c46b7a8b4fdd04b919c812af8292bf648641a37e9fb75ab8b3a5e4bb5f2"
             ],
             "markers": "python_version < '3.14' and python_version >= '3.9'",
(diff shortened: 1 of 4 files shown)


rouille (approved):
Thanks

PR #437, opened 2026-01-22
Use new "steam_load_lbs" column name from pudl

grgmiller:
### Purpose
In PUDL v 2026.1.0, they changed the name of the `steam_load_1000_lbs` column to `steam_load_lbs` (see: https://catalystcoop-pudl.readthedocs.io/en/nightly/release_notes.html#id5). When OGE is installed as a dependency and attempts to read CEMS data from s3, we are currently running into an error since the new `stable` version of the parquet file has this new column name. 

This PR updates the column name and increments the version to 0.7.1. This is intended to be a patch release. 

### What the code is doing
Update column name. Closes OGE-174.

### Testing
Tested locally using GRETA, which uses the oge.load_data function that calls the CEMS data. 

### Where to look
Look in .py files, not notebooks

### Usage Example/Visuals

### Review estimate
5 min

### Future work

### Checklist


rouille (approved):
Looks good

zaneselvans:
Note that we didn't just change the column name, the units also changed (from thousands-of-lbs to lbs) since across the board we try to use the base units (e.g. dollars, not thousands or millions of dollars). Don't know if that has any implications for your use case. I didn't see any calculations in the PR.

grgmiller:
@zaneselvans thanks for following up - we did notice the unit change, but we do not currently use the steam load data in our calculations until we figure out how to translate steam load into MWh where generators report that instead of generation

PR #439, opened 2026-08-06
2025 Early Release Data

grgmiller:
### Purpose
Uses the Early Release version of the 2025 EIA-860 and EIA-923 data to produce an "early release" OGE dataset. The early release versions of these datasets may have some missing, incomplete, or un-validated data. 

The early release caveats:
> This is an early release of the final EIA-860 data for calendar year 2025.  The early release is provided for the express purpose of providing immediate access to individual plant and generator data for analysts who use this type of information.  The data has not been fully edited and is inappropriate for aggregation, such as to state or national totals.  Also, in some cases, data for a certain number of plants and generators has been excluded from this early release pending further data validation. Final, complete, and fully-edited data will be released by EIA later in 2026.  Questions or comments regarding this data release should be directed to eiainfoelectric@eia.gov.

We may not do a full public data release for this due to the warning about aggregation, but this allows us to:
- Update the subplant crosswalk and other intermediate outputs that we use for other purposes
- Get a jumpstart on the final 2025 release that will happen in the fall

Closes DAT-506

### What the code is doing
1. Switches from using `pipenv` as the environment manager to `uv`. This also pins `pandas` to <3 to resolve an issue with our branch of PUDL (this can be unpinned once we switch off pudl). 
2. Updates the `ba_reference` reference table based on updates to the EIA-930 balancing authority metadata, and the FERC allowable entries table (which was updated this year after a long haitus). Part of this process also updates the code in the notebook that is used to update this to automate the process of integrating updates from both sources (AI assisted code change)
3. PUDL updated the column name `wet_dry_bottom` to `wet_dry_bottom_code`; this is now updated in our pipeline.
4. Adds a new helper function to read the raw EIA-860 and 923 early release file names, which do not always follow a strict naming convention (AI assisted code change). If we fully switch to PUDL for these tables, it would obviate the need for this.
5. Adds several new entries to the `epa_eia_crosswalk_manual` file to map EPA units to EIA generators.

### Testing
Successfully ran the entire pipeline for 2025. This uses the stable v2026.7.2 (2026-07-14) of PUDL. 

### Review estimate
20 minutes

### Future work

### Checklist


brandonwang15 (approved):
looks good to me. I also verified that I was able to setup the repository locally and run it locally `uv run python data_pipeline.py --year 2024` (though it hasn't completed yet, as I'm writing this)

PR #443, opened 2026-08-09
Don't output aggregated data for Early Release data runs

grgmiller:
### Purpose
The EIA early release data notes:
> The early release is provided for the express purpose of providing immediate access to individual plant and generator data for analysts who use this type of information. The data has not been fully edited and is inappropriate for aggregation, such as to state or national totals. Also, in some cases, data for a certain number of plants and generators has been excluded from this early release pending further data validation.

I was curious specifically how incomplete the early release data was, and I used Claude to help implement a new analysis notebook to evaluate the data completeness of the monthly and early release versions of 2024 EIA-923 data relative to the final release 2024 data. The top level findings:
- The monthly data only contained about 29% of the records representing about 80-81% of the final generation and fuel data
- The early release data contained about 93% of the records representing 98-99% of the final generation and fuel data

The early release completeness is actually pretty good, so we could potentially debate whether releasing aggregated power sector data would be appropriate. However, our current use case for completing the early release now is primarily for plant-level data, so there's no immediate need to push us to release this data as part of an intermediate data release. 

### What the code is doing
The primary code change is to skip calculation and output of all of the power sector and carbon accounting results if the pipeline is being run for the early release year. 

### Testing
Ran the entire pipeline

### Where to look
`data_pipeline.py`: it is a single `if:` statement, which indents all of the (unchanged) code below it. I also moved the plant attributes table output before the output since we still want to output that, and it was not modified after that point.

### Review estimate
<5 min. 

### Future work
Consider whether early release aggregated results would be appropriate.

### Checklist


src/oge/data_pipeline.py
```diff
@@ -670,154 +637,207 @@ def main(args):
...
-            )[DATA_COLUMNS]
-            .sum(numeric_only=True)
-            .reset_index()
-        )
+        # When running the early release pipeline, we do not want to output aggregated
+        # power sector or carbon accounting data because the plant-level data is
+        # incomplete (analysis of 2024 data revealed that the EIa-923 generation fuel
+        # table is missing about 6-7% of records representing about 2% of the total
+        # generation and fuel).
+        if year <= latest_validated_year:
```
grgmiller, line 646: I don't know why the diff captured it like this, but all I did was add this if statement and then move lines 660-672 to lines 627-639

sarahchute (approved):

PR #448, opened 2026-09-17
Clean up EIA-930 data loading functions

grgmiller:
### Purpose
In preparation for future stacked work, this PR just cleans up the functions in the `eia930.py` file. No code changes are made, this PR just:
- Deletes old functions that are no longer used
- Re-orders functions in our standard order. 

Looking at our current process for loading 930 data, we:
1. Read tables from PUDL
2. Convert new fuel categories to old categories
3. Convert data to gridemissions format
4. Manually adjust timestamps
5. Run physics-based cleaning
6. Re-convert to long format
7. Remove imputed ones

A lot of this code involves just converting the data back and forth so that we can use the physics reconciliation code. We may eventually want to consider in-housing this code to give us more control and avoid the conversion steps, but this is not necessarily the high priority, unless switching to use the new fuel categories will break the physics reconciliation. 

Eventually, we may also want to better automate the manual timestamp adjustment and move this upstream to PUDL (again may be out of scope for current stack),

### Testing
None

### LLM use description
N/A

### Review estimate
5 min - detailed review not required, this just moves functions around.

### Future work
I'll plan to add future stacked PRs on top of this

### Checklist
- [x] Update the documentation (in `docs/docs/`) to reflect changes made in this PR


rouille (approved):
Thanks

